PR #2, opened 2019-01-22
Design stuff

honix:


bakpakin:
Awesome, the doc page navigation looks much better! It was a bit rough around the edges, thanks for the help.

PR #5, opened 2019-02-16
Grammar, spelling and nil and NaN as K/V

maacl:
Description of nil and NaN as keys and values in structs and tables is now consistent with current behavior. Not sure if NaN as value behavior is intended.


bakpakin:
NaN as a value is intended as there is nothing in the runtime implementation that would discourage it. If there is an argument against it, certainly open an issue.

PR #11, opened 2019-02-20
adds more useful information about data structures

xificurC:
I'd like to know the available operations and their complexity on the built-in data structures. Instead of just asking I created a small table and made some wild guesses :) If you agree with the changes you could edit the table and accept the PR.

### DIFF
--- a/md/data_structures.md
+++ b/md/data_structures.md
@@ -69,4 +69,17 @@ itself, and the second parameter is the key.
 (get @"hello, world" 1) # -> 101
 (get "hello, world" 0) # -> 104
 ```
+The table below summarizes the big-O complexity of various operations and other information
+on the built-in data structures.
 
+| Data structure    | Access | Search    | Insert    | Delete    | Space Complexity | Growable? |
+| --------------    | ------ | ------    | ------    | ------    | ---------------- | --------- |
+| Array             | O(1)   | O(n)      | -         | -         | O(1)             | No        |
+| Tuple             | O(1)   | O(n)      | -         | -         | O(1)             | No        |
+| Table             | -      | O(1)      | O(1)      | O(1)      | O(1)             | Yes       |
+| Struct            | -      | O(log(n)) | O(log(n)) | O(log(n)) | O(1)             | Yes       |
+| Buffer            | O(1)   | O(n)      | O(1)      | O(1)      | O(1)             | Yes       |
+| String            | O(1)   | O(n)      | O(log(n)) | O(log(n)) | O(1)             | Yes       |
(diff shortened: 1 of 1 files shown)


bakpakin:
This is a useful chart, but the big(O) is not accurate.

First, all operations on structs are `O(1)`. Structs can also not be modified after creation - neither can tuples or strings. This means appending or prepending to structs, tuples, or strings is linear time `O(n)` relative to the length of the string or number of elements in a struct or tuple, and modifiable (which is probably better stated as "mutable") should be No.

Second, the complexity for keywords and symbols is misleading. They should behave the same way as strings. I would remove that row and put keywords and symbols in with strings.

Third, arrays are indeed growable.

xificurC:
As I noted, the chart was just a mock up with wild guesses. I was too lazy to dig in the sources and thought preparing the chart would save you some time. Note that the guesses were close to my expectations. E.g. since the page says it's clojure inspired I expected structural sharing on structs and tuples. Do you have plans to improve the big(O) complexity of the immutable structures? I think an answer to that question would be a nice addition to this PR as well.

Thank you for your time and your work on programming languages!

bakpakin:
I probably won't add structural sharing by virtue of the fact that `O(1)` reads and cache friendly structures are basically impossible with it. They also make better table keys than other kinds of data structures like red black trees. Many algorithms that take advantage of structural sharing can be more efficiently expressed via mutation, which Janet supports very well.

Techniques like spaghetti stacks, association list, many kinds of trees, etc. usually rely on cons cells to be efficient in lisp, but Janet prioritizes the `get` operation over the `put` operation, and `tuple/prepend` and `tuple/append` are mainly for macro usage and so don't need to be that efficient. They may even be removed, as the `splice` special form makes them mostly unnecessary. (`(tuple/prepend tup 4)` is the same as [;tup 4]). For Janet, if you want efficiency, you should probably use the mutable data structures.

xificurC:
Hm, so immutable data structures are for
- keys in dictionary types (they won't change so the equality check is a pointer check)
- thread-safe shared data

Clojure's approach is different as it trades some algorithmic efficiency for going immutable all the way. In the end it's trade-offs and I understand your choice.

Is there a way to turn a mutable data structure into an immutable one and vice versa? [Pyrsistent](https://github.com/tobgu/pyrsistent) has something called evolvers (never tried actually) that allows you to turn an immutable object into mutable and then back again. With all these questions I'm just pondering on the coding style. The core data structures, their operations, complexity etc. shape the language the most. They decide what is fast, what is idiomatic and what is easy. I'd like to understand that and possibly add it to the docs so others will get it too.

Thanks for your time and the merge!

PR #14, opened 2019-05-10
Fix logo's aspect ratio on mobile

vladimyr:
This is just minor css fix that makes image responsive while preserving its aspect ratio 🎉 

### Before

<hr>


bakpakin:
Thanks. Just curious, what browser and platform are you using? I have tested this on mobile and never seen this.

bakpakin:
Unfortunately, as is, this makes the logo much too small on desktop and larger screens.

vladimyr:
> Thanks. Just curious, what browser and platform are you using? I have tested this on mobile and never seen this.

Chrome 74 on Android 7, which is pretty standard I guess 🤔 

> Unfortunately, as is, this makes the logo much too small on desktop and larger screens.

Actually it does that in Firefox only (which I haven't tested 🤦‍♂ ). Chrom/ium has bug with scaling down images in flex containers ([report](https://bugs.chromium.org/p/chromium/issues/detail?id=625560)) so I failed to notice it. However proper way of doing this would be following:
```css
#logo {
  width: 100% /* tells flexbox to preserve original image dimensions instead of scaling it down (in Firefox too) */
  max-width: 100% /* make sure image is fluid (aspect ratio is preserved: https://alistapart.com/article/fluid-images/) */
  height: auto; /* remove this cause it is default value anyway... */
}
```

vladimyr:
Or `width: 150%;` if that makes you more happy dimension-wise 😉

bakpakin:
Great, I think we had some strange css before which I cleaned up. Setting height: auto seems to be the usually thing to do.

PR #15, opened 2019-05-26
Update doc.html

zaoqi:


bakpakin:
Thanks, this file is auto-generated. I have fixed this upstream and it will be fixed soon on the website.

PR #17, opened 2019-06-09
Fix link to `janet-mode`

dywedir:


bakpakin:
Thank you, my bad for the broken link!

PR #27, opened 2019-09-06
Just the taste of how Grammarly check will look like.

pepe:
Check and improve two paragraphs with Grammarly. Feel free to close it after inspection.

### DIFF
--- a/content/docs/syntax.mdz
+++ b/content/docs/syntax.mdz
@@ -3,22 +3,23 @@
  :order 1}
 ---
 
-A Janet program begins life as a text file, just a sequence of bytes like
-any other on your system. Janet source files should be UTF-8 or ASCII
-encoded. Before Janet can compile or run your program, it must transform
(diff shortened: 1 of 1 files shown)


bakpakin:
Looks better than before, I think using Grammarly on the whole site could definitely improve style. Currently, I usually just use vim's spell checker on pages before deploying. Grammarly seems to like to remove anything in the second person, which is good for essay writing but really a matter of opinion for technical writing. It is best to be consistent, though, and I know I sometimes switch between second person (you should), occasionally first person (we will), and imperative writing.
Feel free to open up other PRs with prose changes.

A nitpick:
- "read time" was changed to "reading time", which is more wrong. It should be "read-time".

pepe:
Thank you! I choose the most formal technical writing (which means no yous, passive voices, stuff like that) on purpose, but let's try some lighter style next time.

PR #35, opened 2020-01-18
fix typo. `new/thread`  shoud be `thread/new`.

btbytes:


bakpakin:
Thanks, keeping docs clean is hard work.

PR #50, opened 2020-05-03
memory-model docs: fix typos; some verbiage tweaks

oakmac:


bakpakin:
Wow, that section was particularly typo-ridden, thanks for the fixes!

PR #52, opened 2020-05-03
destructuring docs: add example output

oakmac:
Adds some code to clarify how the destructing works with a table / struct similar to the other examples on the page.

When I first read this example it was not clear to me what symbols were being bound; I think this helps show the result better.


bakpakin:
Thanks, this is certainly more explicit.

PR #54, opened 2020-05-04
Add lenprefix pattern capture

pepe:
I just copied text from the message of the commit which introduced it.


bakpakin:
Thanks, merging!

PR #57, opened 2020-05-10
Add C source links to the website.

andrewchambers:


andrewchambers:
Note the current Janet version has a bug with os/execute which breaks the build for where-defined.

bakpakin:
Looks good to me, thanks!

PR #58, opened 2020-05-12
configuration: "turned of" -> "turned off"

srenatus:
Hi! Just something that had caught my eye skimming the docs.

I hope editing every version is the right approach here -- is it? 😃


bakpakin:
Do not edit the html files - they are generated. Instead, edit the corresponding mdz file, which is the markup used to generate the site.

srenatus:
Ok cool. I'll update this tomorrow :)

PR #59, opened 2020-05-12
Revert "Add C source links to the website."

bakpakin:
Reverts janet-lang/janet-lang.org#57


bakpakin:
Did not build and the dependency was not included in the repository. Reverting until fixed.

andrewchambers:
Didn't build?

bakpakin:
Yes there were issues, even after the dependency was installed, where the site didn’t seem to build. More to the point though, I would much prefer a small dependency like that to be bundled into the repository.

andrewchambers:
I see, for this use we essentially only need the 'func-info' function from where-defined anyway. The annoying thing is libbacktrace wasn't bundled for my distro.

Issue #69, opened 2020-06-12
Better documentation for comparison (including polymorphic compare)


MikeBeller:
Janet's comparison operators can give some surprising results, especially for comparing values of different types.  Polymorphic comparison was just added to the core in a recent PR.   In closing out issue  janet-lang/janet#272 we discussed adding a chapter to the docs explaining comparison and polymorphic comparison.  I have written a section, and am about to issue a PR to this repo. Just wanted to make sure there was a tracking issue for it before I issued the PR.

MikeBeller:
With the merge of   #70  it closes this issue.

PR #70, opened 2020-06-12
Document comparison in Janet -- including polymorphic comparison

MikeBeller:
Addresses #69


bakpakin:
Thanks, looks thorough.

pepe:
The text is an excellent overview. Thank you.

One thing, that I am not sure about is, that on line 23 and 74 is comparison of the keywords, but in the comment, you are talking about the symbols.

MikeBeller:
Yes @pepe you are right.  My mistake.  I will fix and issue a new PR.  Thanks for catching this.

PR #85, opened 2020-11-13
Tweaks to janet peg sample

sogaiu:
This PR includes a couple of tweaks to the janet peg sample at: https://janet-lang.org/docs/peg.html

1) The first has to do with additional escapes for strings (i.e. `\u....` and `\U......`).  For reference, spork's `fmt.janet` currently has: https://github.com/janet-lang/spork/blob/2e227d4ec8492ee3b51fa7c591e83f126e9e1ad8/spork/fmt.janet#L22-L23

The included change for this just uses newer notation.

2) The other has to do with recognition of things that initially overlap with `:constant` (i.e. `false`, `nil`, `true`).  As an example, `nila` appears to be recognized as two things `nil` and `a`.  Presumably, one would prefer `nila` to be recognized as a single symbol.

May be there's a better way to do this than what's in this PR.

Any ideas?


pepe (approved):
LGTM

PR #93, opened 2020-12-11
Tweak jmp* docs

sogaiu:
Currently the `jmp` docs in the reference table for the abstract machine page say:
```
jmp | (jmp label) | pc = label, pc += offset
```

This was confusing to me because I don't see where `offset` comes from.

I tried using the debugger via the repl to observe the byte code from `disasm` as well as step through to observe actual behavior.

I thought the following alternative more clearly reflected what I saw:
```
jmp | (jmp offset) | pc += offset
```
Does this seem reasonable?

As a concrete example, I started `janet` with `-d` and did:
```
((fn [] (debug) (each i [1] i)))
```
Then I used `(.step) (.ppasm)` to step through instructions and observe their effects:
```
  signal: 
  function:    [repl]
  constants:  @[]
  slots:      @[nil 1 (1) nil 1]

   ldn 1                # line 1, column 9
   sig 0 1 2           
   ldi 1 1              # line 1, column 25
   push 1              
   mktup 1             
   movn 2 1             # line 1, column 17
   ldn 3               
   next 1 2 3          
   movn 3 1            
   jmpni 3 5           
   in 1 2 3            
   movn 4 1            
   next 3 2 3          
 > jmp -4              
   retn
```

This PR reflects similar changes to the other `jmp*` instructions as well.


pepe (approved):
LGTM

bakpakin:
That's a good point, in most cases when looking at bytecodebjmp will be an offset. However you can also use keyword labels so that you don't need to manually calculate the offset.

PR #100, opened 2021-01-25
Cover both `try` and `protect` in error section

subsetpark:
Adds a section on the error handling page about `protect`. `protect` and `try` seem to be equally "first-class" in terms of conveniences exposed around fiber errors, so let's document both.

### DIFF
--- a/content/docs/fibers/error_handling.mdz
+++ b/content/docs/fibers/error_handling.mdz
@@ -32,8 +32,14 @@ from the last call to @code`resume`.
  (print "value returned: " res))
 ```
 
-Janet also provides a simple macro @code`try` to make this a bit easier in the
-common case.
+## Two error-handling forms built on fibers
+
+### @code`try`
+
+Janet provides a simple macro @code`try` to make error-handling a bit
+easier in the common case. @code`try` performs the function of a
+traditional try/catch block; it will execute its body and, if any
+error is raised, pattern match against the error in its second clause.
 
 @codeblock[janet]```
 (try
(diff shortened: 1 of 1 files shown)


content/docs/fibers/error_handling.mdz
```diff
@@ -32,8 +32,14 @@ from the last call to @code`resume`.
...
-Janet also provides a simple macro @code`try` to make this a bit easier in the
-common case.
+## Two error-handling forms built on fibers
+
+### @code`try`
+
+Janet provides a simple macro @code`try` to make error-handling a bit
+easier in the common case. @code`try` performs the function of a
+traditional try/catch block; it will execute its body and, if any
+error is raised, pattern match against the error in its second clause.
```
pepe, line 42: I do not think it is pattern matched. Can you please elaborate on the idea? When looking at https://github.com/janet-lang/janet/blob/master/src/boot/boot.janet#L232, I do not see any pattern matching.

content/docs/fibers/error_handling.mdz
```diff
@@ -43,3 +49,16 @@ common case.
    (print "will never get here"))
  ([err] (print "caught error: " err)))
 ```
+
+### @code`protect`
+
+Janet also provides the @code`protect` macro for a slightly different
+flavor of error-handling, converting caught errors into further
```
pepe, line 56: I would go with something like "returning the tuple of two members, [true result] or [false error]" but in your fine language tho. And the nice example I see here :).

content/docs/fibers/error_handling.mdz
bakpakin: There is indeed no pattern matching - there is just two optional bindings - the error itself and the fiber that raised the error.

subsetpark:
@bakpakin Removed phrase "pattern match"

pepe (approved):
LGTM

PR #103, opened 2021-01-30
Update C API docs on importing native module

matthewmcgarvey:
Without the leading slash it just looks in janet's sources instead of the local directory and failss


bakpakin:
Thanks, this change was fairly recent so the docs are outdated.

PR #104, opened 2021-02-08
Add optional tag arg to line and column docs

sogaiu:
This seems to work:
```
repl:1:> (peg/match ~(sequence "a" (line :a) (backref :a)) "a")
@[1 1]
```
If that's correct behavior (as well as for `column`) may be the table entries for them can be updated to have the `?tag` argument.


pepe (approved):
LGTM.

Can you please add the sentence that both these captures do not advance? Similar to the position capture. I added these two and forgot this :facepalm:

pepe:
Nevermind, I will do it myself later.

PR #106, opened 2021-02-13
Use :in instead of :generate in loop example

sogaiu:
Update loop example to use `:in` instead of `:generate`.

Thanks to @Alligator for reporting.


pepe (approved):
LGTM

PR #108, opened 2021-02-20
Change non-existent function 'interpolate' to 'interpose' in jpm docs

SteadBytes:
The `jpm` docs contain an example to print command-line arguments on a single line, separated by commas. A function `interpolate` is used which (as far as I can tell) does not exist. This PR changes the example to use `interpose` which produces the same result.


sogaiu:
LGTM

PR #110, opened 2021-02-27
Fix absolute path in the example

oz123:
The code without this fix will result in an error:

```
repl:1:> (import /build/mymod :as mymod)
error: could not find module /build/mymod:
    /usr/local/lib/janet/build/mymod.jimage
    /usr/local/lib/janet/build/mymod.janet
    /usr/local/lib/janet/build/mymod/init.janet
    /usr/local/lib/janet/build/mymod.so
    /build/mymod.jimage
    /build/mymod.janet
    /build/mymod/init.janet
    /build/mymod.so
  in require-1 [boot.janet] on line 2774, column 20
  in import* [boot.janet] on line 2813, column 15
  in _thunk [repl] (tailcall) on line 1, column 1
```

Not a big deal for an experienced Linux user I guess. Never the
less, we all do copy paste and expect things to just work :-).

### DIFF
--- a/content/capi/index.mdz
+++ b/content/capi/index.mdz
@@ -93,7 +93,7 @@ If all goes well, @code`jpm` should have built a file @code`build/mymod.so`. In 
 import your module and use it.
 
 @codeblock[janet]```
-(import /build/mymod :as mymod)
+(import ./build/mymod :as mymod)
 (mymod/myfun) # Prints the hello message
 ```
 


sogaiu:
Hmm, I think in recent versions either may work if janet is started from one's project directory.

From the [CHANGELOG](https://github.com/janet-lang/janet/blob/master/CHANGELOG.md#1141---2021-01-18):

> Importing modules that start with / is now the only way to import from project root. Before, this would import from / on disk. Previous imports that did not start with . or / are now unambiguously importing from the syspath, instead of checking both the syspath and the project root. This is backwards incompatible and dependencies should be updated for this.

What version are you testing with?

From a project directory, either `(import /build/...)` or `(import ./build/...)` worked here on a Linux on https://github.com/janet-lang/janet/commit/d6ea1989cca4d44b47d14d50dd3cbed31c392fbf.

```
repl:1:> (import ./build/pcgrand :as p)
@{p/make @{:private true} _ @{:value <cycle 0>} p/random @{:private true} p/srandom @{:private true} p/boundedrand @{:private true}}
repl:2:> 

$ janet
Janet 1.15.2-local linux/x64 - '(doc)' for help
repl:1:> (import /build/pcgrand :as p)
@{p/make @{:private true} _ @{:value <cycle 0>} p/random @{:private true} p/srandom @{:private true} p/boundedrand @{:private true}}
```

That was a test with: https://github.com/sogaiu/janet-pcg-random

oz123:
Oh. Darn. I didn't see that coming 🤦.
I use 1.13.

PR #112, opened 2021-03-08
Fix error in memory management sample code

fimmtiu:
The arguments to `janet_tuple_n` are `(const Janet *values, int32_t n)`, according to https://github.com/janet-lang/janet/blob/5c05dec65ae4b49e65b01f96984b81c8ca774ad4/src/core/tuple.c#L50, so the second argument to it in this sample code should be `argc` rather than `argv`.


sogaiu:
Looks like an improvement to me.

pepe (approved):
LGTM

PR #123, opened 2021-07-29
Modify the values of `set` special form to be differently with ones of `put`

soulawaker:
Modify the values of `set` special form to be different with ones of `put`

### DIFF
--- a/content/docs/data_structures/buffers.mdz
+++ b/content/docs/data_structures/buffers.mdz
@@ -78,8 +78,8 @@ The @code[set] special form also works for setting bytes in a buffer.
 
 @codeblock[janet](```
 (def b @"")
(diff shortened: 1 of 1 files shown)


sogaiu:
It's true that the return values as expressed don't make sense so the suggested changes are more accurate.

Another alternative modification is something like:
```
(set (b 0) 97) # b is now @"a"
(set (b 10) 97) # the content of b is now @"a\0\0\0\0\0\0\0\0\0a"
```

uvtc:
Note for consistency with the rest of the docs, I'd suggest also changing any occurrences of `#->` to `# ->`.

bakpakin:
Yup, looks correct now.

PR #125, opened 2021-08-21
installing from source, add instruction for jpm

uvtc:


pepe (approved):
Lgtm

sogaiu:
The text above the jpm part talks briefly about customizing installation:

> This will install in @code[/usr/local] by default, see the Makefile to customize.

Possibly worth a word or two that something similar is possible with jpm?

Not sure how to say it briefly but what I'm referring to is the sort of latter half of: https://github.com/janet-lang/jpm/#bootstrapping

uvtc:
In that sentence under [macOS and Unix-like](https://janet-lang.org/docs/index.html#macOS-and-Unix-like), where it says "see the Makefile to customize", does that mean doing something like:

    PREFIX=/path/to/somewhere make
    PREFIX=/path/to/somewhere make test
    PREFIX=/path/to/somewhere make install

? The docs aren't suggesting modifying the makefile, are they?

Also, I don't understand the text in [the jpm readme](https://github.com/janet-lang/jpm/blob/master/README.md) about using those config files. I installed onto Debian GNU/Linux and didn't need to do

    janet cli.janet install --bootstrap-config=configs/linux_config.janet

I just did the plain

    janet cli.janet install

and it worked fine. What is the purpose of the config files?

Finally, still regarding installing jpm, my first/wild guess would be that, since it's *after* I have Janet installed, if I wanted to alter something, I'd use one of the Janet env vars, for example:

    JANET_PATH=/path/to/somewhere janet cli.janet install

So, I think I'm missing some understanding here. Any clarification much appreciated.

pepe:
@uvtc I guess this https://github.com/janet-lang/jpm/blob/master/default-config.janet is the answer, why you did not need to provide the config

PR #126, opened 2021-08-21
update repl snippet for v1.17.0

uvtc:


uvtc:
Maybe not readily apparent from the diff, this was just a copy/paste from my terminal, running the same commands as in the original repl snippet.

PR #116, opened 2021-04-12
Document `:install true` for `declare-executable`

sogaiu:
This is an attempt to document the `:install true` key-value pair for the `declare-executable` form in `project.janet`.

PR #134, opened 2021-09-16
Added section for "optional flags"

uvtc:


sogaiu:
@uvtc I looked into the checks failing for this PR.

I don't know the reason, but it builds here after I remove the double quotes (6 of them) from these lines: https://github.com/janet-lang/janet-lang.org/blob/02e6284ea5f9f7fa264f439af743fc6f0e74b459/content/docs/functions.mdz#L265-L266

I also had a PR that wouldn't build (you can check by [running `make` locally](https://github.com/janet-lang/janet-lang.org/pull/116#issuecomment-967766507)) but @pepe gave me a good hint to fix the issue I had.

Here it looks like it might be something different.

uvtc:
Thanks, @sogaiu !

Issue #137, opened 2021-10-13
Link to Joy 404s


elimisteve:
At https://janet-lang.org/ , the link to "Joy Web Framework" leads to joyframework.com , which returns a 404.

elimisteve:
@oakmac Yup, that's what my PR does, and what @swlkr says is correct: https://github.com/janet-lang/janet-lang.org/pull/138

sogaiu:
@elimisteve Since #138 was merged (the link on the site now works for me too), may be this issue can be closed?

elimisteve:
@sogaiu Yes indeed!

PR #138, opened 2021-10-13
index.mdz: Update link to Joy Web Framework

elimisteve:


elimisteve:
Fixes https://github.com/janet-lang/janet-lang.org/issues/137

PR #143, opened 2021-10-19
Tweaks to event loop docs

sogaiu:
Mostly minor tweaks including some code indentation changes.

Includes some removal of what I thought were references to older thread functionality.

### DIFF
--- a/content/docs/event_loop.mdz
+++ b/content/docs/event_loop.mdz
@@ -3,7 +3,7 @@
  :order 21}
 ---
 
-In addition to threads, Janet comes with another powerful concurrency model out of the box - the event loop. The event loop provides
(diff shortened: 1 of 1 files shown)


pepe:
Also, I would think that ev should come before the thread article.

sogaiu:
Ah, may be that would make more sense.

I guess changing the following numbers appropriately might work?

* https://github.com/janet-lang/janet-lang.org/blob/master/content/docs/networking.mdz#L3
* https://github.com/janet-lang/janet-lang.org/blob/master/content/docs/threads.mdz#L3
* https://github.com/janet-lang/janet-lang.org/blob/master/content/docs/networking.mdz#L3

If the idea makes sense, may be it would be better to do it after decisions about some of the existing PRs are made.

PR #149, opened 2021-11-26
Update index.mdz

uvtc:
I think this wording is more clear, since they are not exactly interchangeable, and "interchangeable" suggests something like the interchangeability in Racket (IIUC).


sogaiu:
I agree the current wording could lead to confusion.

PR #157, opened 2022-03-22
Tweak Linux install info

sogaiu:
@levitanong reported on the janet-language/help channel about experiences with using the releases `.tar.gz` for Linux.

bakpakin provided clarification regarding the intention of that `.tar.gz`:

> So there seems to be significant discussion about installing Janet releases using the Tarball on linux - my advice there is use the build instructions and build from source instead. The tarball is mainly for people scripting releases - and this is the case for most software on various unixes, simply because it is expected that distros will link to different libc versions, turn on different optimization flags, etc.

This PR is an attempt to tweak the existing installation docs to better reflect the above information.

### DIFF
--- a/content/docs/index.mdz
+++ b/content/docs/index.mdz
@@ -6,10 +6,8 @@
 
 ## Installation
 
-Install a stable version of Janet from the
-@link[https://github.com/janet-lang/janet/releases][releases page].  Janet is
-prebuilt for a few systems, but if you want to develop Janet, run Janet on a
-non-x86 system, or get the latest, you must build Janet from source.
+Janet is prebuilt for a few systems, but if you want to develop Janet, run Janet
+on a non-x86 system, or get the latest, you must build Janet from source.
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


levitanong:
Would it also help to explicitly say the tarball is meant for people scripting releases?

levitanong:
I also think bakpakin's example would be helpful to add somehow, to illustrate the idiomatic way to customize installation.

https://gitter.im/janet-language/help?at=623923e7161ffc40d7c7f338

sogaiu:
> Would it also help to explicitly say the tarball is meant for people scripting releases?

I am not sure.  I am not opposed to the idea, but don't really have a good idea of how to execute it well.

Thus I prefer not to add clarification about the Linux releases tarball to this PR, but I have no problem with someone else making a separate PR for that.

On a side note, I think that adding text on the page in question will not necessarily help for folks who go directly to the releases page.

uvtc:
> On a side note, I think that adding text on the page in question will not necessarily help for folks who go directly to the releases page.

Yeah, if you go to the main Janet webpage, and click downloads, you're taken straight to the releases page at github where it's just links to download files (no instructions on what to do there). I bet it's tempting for a new prospective user on GNU/Linux to click the link with "linux" in it rather than the source.

A full solution might be to:

  * change docs/index.html into docs/install.html, and have index.html only contain the "first program" section already there
  * at janet-lang.org, change the top nav "Download" link to point to docs/install.html

bakpakin:
Yeah, this is good clarification.

PR #162, opened 2022-04-20
double-quoted strings can span multiple lines

uvtc:


sogaiu:
For the multiline string example:
```
"Hello, this
    is all
one line
here."   # -> "Hello, this    is allone linehere."
```
have you considered something like:
```
"Hello, this
    is al
l one lin
e here."   # -> "Hello, this    is all one line here."
```
?

uvtc:
Thanks, @sogaiu . putting the "al" and the "l" on separate lines more clearly shows how they get joined together (and the newline is removed). I kept the "linehere" as-is because this way the example shows three cases all in one string: the 4 spaces preserved, the two word fragments forming a word, and the two words glued together.

pepe (approved):
LGTM

PR #166, opened 2022-11-09
fixed example for named functions

tionis:
I just realized that my last pull request added the quotation marks in the wrong place for the example to make sense
Also wording.


pepe (approved):
LGTM

PR #176, opened 2023-03-09
Fix typos in ffi/write and ffi/read code blocks

CFiggers:
Currently, the sample code blocks on the Foreign Function Interface page contain invalid syntax:

image

image

These snippets do not compile because the `@` is inside the `[`:

```
Janet 1.27.0-01aab666 linux/x64/gcc - '(doc)' for help
repl:1:> (def my-type (ffi/struct :char [@:int 4]))
repl:1:32: compile error: unknown symbol @:int
```

This PR fixes the sample code by moving the `@` outside the `[` in both code blocks:

```
Janet 1.27.0-01aab666 linux/x64/gcc - '(doc)' for help
repl:1:> (def my-type (ffi/struct :char @[:int 4]))
<core/ffi-struct 0x563C9B5EE910>
```

### DIFF
--- a/content/docs/ffi.mdz
+++ b/content/docs/ffi.mdz
@@ -133,7 +133,7 @@ a @code`:ptr` type to a function.
 
 @codeblock[janet]```
 (ffi/context "./mylib.so")
-(def my-type (ffi/struct :char [@:int 4]))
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #177, opened 2023-03-31
don't autofocus the search bar

### LINKED ISSUE #170: Auto-focusing search means you can't link to specific API functions
If I follow a link like this:

https://janet-lang.org/api/index.html#map

I would expect it to take me directly to the `map` function. And in fact it does exactly that -- but if your browser executes JavaScript, the little search widget thing will scroll the page all the way up to the top a moment later.

ianthehenry:
This prevents sharing links to individual functions using the URL fragment, as autofocusing the searchbox will scroll to the top of the page.

Fixes #170

### DIFF
--- a/templates/docpage.html
+++ b/templates/docpage.html
@@ -22,7 +22,7 @@
       <div class="bar"></div>
       <div class="bar botbar"></div>
     </div>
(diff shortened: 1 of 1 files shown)


sogaiu:
The change seems to be working here -- I tried out the change via a locally running server (`make run`).

Tested with both FF and Brave.

Thanks!

PR #178, opened 2023-04-01
Incorrect function referenced in `(product)` example

CFiggers:
The example for the `(product)` function contains a reference to `(sum)`, but functioning in context the way one would expect `(product)` to function. This was potentially introduced as a copy/paste error.

image


pepe (approved):
LGTM

PR #180, opened 2023-05-07
Update supervisor event description

sogaiu:
In my local testing, I'm observing supervisor events that are tuples with 3 elements, though the current website docs describe such events as being tuples of two elements.  Perhaps things changed at some point (or may be I'm doing something wrong (^^; )

Also, the keyword that's the first element of the tuple seems to currently be `:ok` and not `:done` for the non-error case.

This PR is an attempt to reflect these findings in the docs.

I've tried to describe the third element of the tuple based on code from [here](https://github.com/janet-lang/janet/blob/cabbaded68947a7be3bee5c5b9e0dff0a3963c48/src/core/ev.c#L548-L552).  Happy to change to a better description :)

### DIFF
--- a/content/docs/threads.mdz
+++ b/content/docs/threads.mdz
@@ -105,7 +105,7 @@ Thread supervisors need to be specified when creating the thread.
 (ev/thread worker nil :n supervisor)
 
 # Get one event from the supervisor channel (on the initial thread here)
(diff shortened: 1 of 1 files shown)


bakpakin:
LGTM - yes the task-id is the new element to make it easier to distinguish where an event came from without hacking the payload (in the threaded case specifically).

PR #184, opened 2023-05-19
small change to API layouting; better use of vertical screen space

zevv:


zevv:
Example available at http://zevv.nl/janet-docs/api/array.html

sogaiu:
[Here](https://janet-lang.org/api/array.html) is a link to the original for comparison purposes.

PR #188, opened 2023-06-17
Tweak abstract machine reference table

sogaiu:
This PR contains some suggestions for the abstract machine reference table.

Specifically the changes are:

* Added missing `dest` argument for `prop`
* Swapped the last two arguments of `puti`
* Changed spelling of `tcheck` to `tchck`

On a side note, regarding the argument swapping for `puti` -- the order was surprising to me given the pattern for `get`, `geti` and `put`.  I looked a bit at some history for it and it seems to have been that way for a long time.  Curious about the order.

Any hints?

### DIFF
--- a/content/docs/gen-asm-table.janet
+++ b/content/docs/gen-asm-table.janet
@@ -70,7 +70,7 @@
    ['res '(res dest fiber val) "$dest = resume $fiber with $val"]
    ['ret '(ret val) "Return $val"]
    ['retn '(retn) "Return nil"]
-   ['setu '(setu env index val) "envs[env][index] = $val"]
+   ['setu '(setu val env index) "envs[env][index] = $val"]
    ['sig '(sig dest value sigtype) "$dest = emit $value as sigtype"]
    ['sl '(sl dest lhs rhs) "$dest = $lhs << $rhs"]
    ['slim '(slim dest lhs shamt) "$dest = $lhs << shamt"]
--- a/content/docs/gen-asm-table.janet
+++ b/content/docs/gen-asm-table.janet
@@ -59,13 +59,13 @@
    ['neqim '(neqim dest lhs im) "$dest = $lhs != $im"]
    ['next '(next dest ds key) "$dest = next($ds, $key)"]
    ['noop '(noop) "Does nothing."]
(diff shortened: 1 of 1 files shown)


sogaiu:
I'm still trying to clarify my understanding of `ldu` and `setu`, but is it possible that the info in the table is slightly off for `setu` as well?

Currently, `vm.c` has the following for each:

```c
    VM_OP(JOP_LOAD_UPVALUE) {
        int32_t eindex = B;
        int32_t vindex = C;
        JanetFuncEnv *env;
        vm_assert(func->def->environments_length > eindex, "invalid upvalue environment");
        env = func->envs[eindex];
        vm_assert(env->length > vindex, "invalid upvalue index");
        vm_assert(janet_env_valid(env), "invalid upvalue environment");
        if (env->offset > 0) {
            /* On stack */
            stack[A] = env->as.fiber->data[env->offset + vindex];
        } else {
            /* Off stack */
            stack[A] = env->as.values[vindex];
        }
        vm_pcnext();
    }

    VM_OP(JOP_SET_UPVALUE) {
        int32_t eindex = B;
        int32_t vindex = C;
        JanetFuncEnv *env;
        vm_assert(func->def->environments_length > eindex, "invalid upvalue environment");
        env = func->envs[eindex];
        vm_assert(env->length > vindex, "invalid upvalue index");
        vm_assert(janet_env_valid(env), "invalid upvalue environment");
        if (env->offset > 0) {
            env->as.fiber->data[env->offset + vindex] = stack[A];
        } else {
            env->as.values[vindex] = stack[A];
        }
        vm_pcnext();
    }
```

That suggests to me that the info in the table (the middle column):

```
ldu | (ldu dest env index) | $dest = envs[env][index]
```

and:

```
setu | (setu env index val) | envs[env][index] = $val
```

should be more symmetric.  That is, `setu` should be called like `(setu val env index)`.

sogaiu:
My current understanding is that `setu`'s signature is `(setu val env index)` so the table's current content could use some adjustment.

Thus, the additional change in [c779fdd](https://github.com/janet-lang/janet-lang.org/pull/188/commits/c779fddc71dae05c436a5a1ba072b70ade227d0e) seems necessary to me.

PR #190, opened 2023-06-23
Fix a small typo in the documentation for match

mwilbur:
I'm making an assumption that "Easy case" should be "Each case".  I couldn't see anywhere the html is auto-generated for this so I changed the html itself.


sogaiu:
I believe the API docs are generated based on the docstrings from the janet source code, may be using [this code](https://github.com/janet-lang/janet-lang.org/blob/604b786dac163ce68f0eba7c863e5a7bb879745f/content/api/gen-docs.janet).

I think it's true that "Easy case" should be "Each case".  However, [the current docstring does say "Each case"](https://github.com/janet-lang/janet/blob/c3f4dc0c152dd1d08fdd96e406a2847e1a3f9b40/src/boot/boot.janet#L1792).  This was fixed in https://github.com/sogaiu/janet/commit/c864828735bef98e5b9a18015167133b75ee5662.  This might explain why more recent docs say "Each case".

I think there might be value in being able to see the docs as they were at the time they were generated.  I don't know what the policy is though.

mwilbur:
Makes sense to me.

PR #199, opened 2023-07-18
Fix FFI section: corresponding C type name

iacore:


sogaiu:
I don't understand the details very well, but I'm curious why `ptrdiff_t` and not `ssize_t`.

sogaiu:
That sounds similar to the following from [the mailing list post](https://lists.gnu.org/archive/html/emacs-devel/2014-10/msg00019.html):

> 3. ptrdiff_t is more ubiquitous and better-standardized than ssize_t, as ptrdiff_t is required by the C standard whereas ssize_t is required only by POSIX.

I wonder if it doesn't make sense for `ffi.c` to use `:ptrdiff` instead of `:ssize` then...

```
@tr{@td{@code`:ptrdiff`} @td{ptrdiff_t}}
```

iacore:
> I wonder if it doesn't make sense for `ffi.c` to use `:ptrdiff` instead of `:ssize` then...
> 
> ```
> @tr{@td{@code`:ptrdiff`} @td{ptrdiff_t}}
> ```

We can support both `:ptrdiff` and `:ssize` (backwards compatibility).

sogaiu:
Ah, there's an idea :)

sogaiu:
> Question: why is float r32 not f32?

Wondered the same.

Looking at my QWERTY keyboard, I see `r` is quite close to `f`...but then I came across [this](https://michael-f-bryan.github.io/scad-rs/noisy_float/types/index.html).

Issue #198, opened 2023-07-18
Extra dots in some operator symbols in ref table for abstract machine instructions?


sogaiu:
There are some operators in the description column of [the reference table for the abstract machine's intructions](https://janet-lang.org/1.29.1/docs/abstract_machine.html#Reference-table) that start with a dot:

```
$dest = $lhs .>= $rhs
$dest = $lhs .> im
$dest = $lhs .<= $rhs
$dest = $lhs .< im
```

May be these dots are not intentional?

Issue #196, opened 2023-07-12
Bytecode reference table lacks entry for subim


sogaiu:
At the time of this writing, the [reference table for bytecodes](https://janet-lang.org/1.29.1/docs/abstract_machine.html#Reference-table) lacks a row for the new bytecode instruction `subim`, added in https://github.com/janet-lang/janet/commit/b5d3c8725335d18c69e84c4a01adb7ef27a5d473 .

May be adding something like the following between [these two rows](https://github.com/janet-lang/janet-lang.org/blob/604b786dac163ce68f0eba7c863e5a7bb879745f/content/docs/gen-asm-table.janet#L81-L82):

```janet
 ['subim '(subim dest lhs im) "$dest = $lhs - im"]
```

is sufficient to address this issue.

Issue #194, opened 2023-07-01
Bytecode reference table lacks entry for divf


sogaiu:
At the time of this writing, the [reference table for bytecodes](https://janet-lang.org/1.29.1/docs/abstract_machine.html#Reference-table) lacks a row for the new bytecode instruction, `divf`, added in https://github.com/janet-lang/janet/commit/c83f3ec09757eb48bf7d48f3063b39e4d8bd9345.  

Perhaps inserting an appropriate row between [these two rows](https://github.com/janet-lang/janet-lang.org/blob/604b786dac163ce68f0eba7c863e5a7bb879745f/content/docs/gen-asm-table.janet#L19-L20):

```
   ['div '(div dest lhs rhs) "$dest = $lhs / $rhs"]
   ['divim '(divim dest lhs im) "$dest = $lhs / im"]
```

is sufficient to address this issue.

I wonder what to use to indicate the operator in the rightmost column:

```
 ['divf '(divf dest lhs rhs) "$dest = $lhs ??? $rhs"]
```

One candidate for `???` might be `div` because:

```
$ janet
Janet 1.29.1-58d29736 linux/x64/gcc - '(doc)' for help
repl:1:> (disasm (fn [] (div 2 3)) :bytecode)
@[(ldi 1 2) (ldi 2 3) (divf 0 1 2) (ret 0)]
```

However, this might be confusing because the row above it would presumably be this:

```
['div '(div dest lhs rhs) "$dest = $lhs / $rhs"]
```

That is, the instruction's name in that row is `div`.

May be using `//` makes sense?

sogaiu:
In support of using `//`, in Python, it appears that:

```
$ python
Python 2.7.18 (default, Jul  1 2022, 10:30:50) 
[GCC 11.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> -7 // 5
-2
```

..and with a recent janet (see earlier comment for info about bytecode being `divf`), we have:

```
$ janet
Janet 1.29.1-db0abfde linux/x64/gcc - '(doc)' for help
repl:1:> (div -7 5)
-2
```

---

On a terminological note, there appears to be a lack of consensus about what "integer division" refers to:

[...]

PR #200, opened 2023-07-23
Tweak abstract machine reference table - address #194, #196, and #198

sogaiu:
This PR is an attempt to address #194, #196, and #198.

The changes consist of:

* Adding rows for two new bytecodes
* Removing extra dots from in front of some operators


pepe (approved):
LGTM

PR #203, opened 2023-12-16
Document peg grammar: 0

iacore:
reference: https://github.com/janet-lang/janet/pull/1187#issuecomment-1586490024

### DIFF
--- a/content/docs/peg.mdz
+++ b/content/docs/peg.mdz
@@ -82,7 +82,9 @@ languages.
         @td{ The literal string. }}
 
     @tr{@td{3 (an integer)}
-        @td{ Matches a number of characters, and advances that many characters.
+        @td{ 
+		If positive, matches a number of characters, and advances that many characters.
+		If 0, counts as an (empty string) match.
         If negative, matches if not that many characters and does not advance.
         For example, -1 will match the end of a string. }}
 


sogaiu:
For comparison, the following is what [janet-pegdoc](https://github.com/sogaiu/janet-pegdoc) says about integer peg special patterns:

```
$ pdoc -d integer

`<n>` -- where <n> is an integer

For n >= 0, try to match n characters, and if successful, advance that many characters.

For n < 0, matches only if there aren't |n| characters, and do not advance.

For example, -1 will match the end of a string because the length of the empty string is 0, which is less than 1 (i.e. |-1| = 1 and there aren't 
that many characters).
```

iacore:
I changed the wording a bit. It still reads awkward but the meaning is clear.

sogaiu:
I think that this line:

```
For example, -1 will match the end of a string.
```

should come right after this line:

```
If negative, matches if not that many characters and does not advance.
```

because `-1` is an example of a negative number, while `0` is not.

I think it would be fine to address the case of `0` as the last line (or before the text having to do with negative integers), for example:

```
0 counts as a match but there is no advancing.
```

iacore:
I changed the wording again

sogaiu:
The latest looks much improved to me :)

PR #204, opened 2024-01-01
document the (sub) PEG special

ianthehenry:
What it says on the tin!


sogaiu:
Thanks for submitting this :+1: 

Possibly slightly clearer?

```
If @code`patt` also succeeds, @code`sub` will
advance the input to the end of what @code`window-patt`
matched.
```

or may be:

```
If @code`patt` also succeeds, @code`sub` will
advance the input to the end of the bytes that
@code`window-patt` matched.
```

?

ianthehenry:
I think that is better -- I tweaked the language there to be basically your first example.

PR #210, opened 2024-02-01
Mention fiber/status + last-status in main text

sogaiu:
Following on from the point made in [this comment](https://github.com/janet-lang/janet/issues/1365#issuecomment-1919470197), this PR is an attempt to appropriately mention `fiber/status` and `fiber/last-value` in the main text.

Note that although `fiber/status` appears in a code example, it doesn't appear to be explained in the main text.

If I've missed relevant mentions of these, please point them out.  I know they appear in the multithreading page, but there they are mentioned in passing in a manner that doesn't introduce them to folks who might be unfamiliar with them.


sogaiu:
Thanks for the suggestion, I will consider it.

---

If you look over the histories of various Janet repositories, I think you'll see that I (and others) have been motivated over the years to improve various things including documentation.

This has been happening without your llama comments.  Those are unnecessary, distracting, and frankly likely to lead to annoyance.  If you persist in that activity, I think it may lead to a decrease in motivation and/or redirection of time, energy, and effort in counterproductive or disspitative things (e.g. I am not spending time improving Janet documentation here because I'm writing this part of the response).  That seems contrary to improving Janet so I suggest you stop.

amano-kenji:
Right. You don't need a llama. I'm going to stop being a llama.

sogaiu:
I've made some more changes and pushed them to this PR, both to the Fiber overview page and the Error Handling page.

I'm likely to be mulling over the content more, but am not adverse to hearing about the content in its current state (^^;

amano-kenji:
I finished reading it. It is excellent.

sogaiu:
Thanks for taking a look.

---

I've added a note about synonyms for the term "trap" in reference to signals.  The ones I've come across so far include:

* "block" - `fiber/new`'s docstring
* "capture" - `fiber/new`'s docstring
* "mask" - `with-dyns`' docstring

I didn't find "trap" in any docstring [1], but it occurs on the [Fibers overview page](https://janet-lang.org/docs/fibers/index.html).

---

[1] To do this kind of search, it's convenient to search on [this page](https://janet-lang.org/api/index.html).  Though when the docs are not for the latest version of Janet, the search is perhaps not quite complete.

bakpakin:
Just for some explanation of why it's missing, fibers didn't use to store their last values at all and in many cases it is redundant. However it was needed for a couple of things like iterating over fibers and the certain things in the ev module.

sogaiu:
Thanks for the background!

Issue #209, opened 2024-01-26
Document `declare-bin`.


amano-kenji:
I noticed that https://janet-lang.org/docs/jpm.html doesn't mention `declare-bin`.

sogaiu:
This looks similar to https://github.com/janet-lang/janet-lang.org/issues/34.

sogaiu:
There is now a PR (#212) with a proposal for an additional section that includes some documentation for some `declare-` callables, including `declare-bin`.

amano-kenji:
:cat2:

sogaiu:
As #212 has been merged, I think this can be closed now.

Issue #34, opened 2020-01-06
jpm declare-binscript is not documented on the website.


andrewchambers:

sogaiu:
Until there are more docs, [jpm's declare.janet](https://github.com/janet-lang/jpm/blob/master/jpm/declare.janet) might be worth taking a look at.  It has `defn` forms for:

* declare-native
* declare-source
* declare-headers
* declare-bin
* declare-executable
* declare-binscript
* declare-archive
* declare-manpage
* declare-project

...and all of them have docstrings.

sogaiu:
The current state of things appear to be that of the `declare-` contructs mentioned above, the following seem to lack documentation on the website:

* declare-archive
* declare-bin <- #209
* declare-binscript <- #34
* declare-executable
* declare-headers
* declare-manpage

For reference, `declare-binscript`'s docstring is currently:

> Declare a janet file to be installed as an executable script. Creates a shim on windows. If hardcode is true, will insert code into the script such that it will run correctly even when JANET_PATH is changed. if auto-shebang is truthy, will also automatically insert a correct shebang line.

sogaiu:
Closing as #212 has been merged.

sogaiu:
As #212 has been merged, I think this one can be closed now.

That doesn't seem to be something I can do though.

PR #212, opened 2024-02-03
Add some jpm declare- docs (#34, #209)

sogaiu:
This PR is an attempt to address #34 and #209.

It adds a new section to the [jpm](https://janet-lang.org/docs/jpm.html) page that contains some docs for the following `declare-` callables:

* `declare-bin`
* `declare-binscript`
* `declare-headers`
* `declare-manpage`

I have seen each of these used in at least one project [1], so it seemed worth doing.

I based the added docs on the docstrings found in `jpm`'s `declare.janet`.  I made some changes though because some things didn't seem right or there seemed to be missing bits.

Also snuck in a minor formatting change to a code example.  I hope that can be forgiven (^^;

---

[1] I did not add a docstring for one other construct I found in jpm's `declare.janet` because I have not encountered a single project that uses it.

### DIFF
--- a/content/docs/jpm.mdz
+++ b/content/docs/jpm.mdz
@@ -248,11 +248,12 @@ the entry point to your executable.
 # This will be printed when you run `jpm build`
 (print "build time!")
 
-(defn main [& args]
-    # You can also get command-line arguments through (dyn :args)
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

Issue #205, opened 2024-01-20
Document `&keys` better.


amano-kenji:
Until recently, I didn't recognize the example for `&keys` on https://janet-lang.org/docs/functions.html was using destructuring. I thought the destructured struct was the only allowed syntax for  `&keys`.

I didn't think it was a destructured struct. I thought it was just the only syntax for `&keys`.

```
(defn test
  [&keys {:argument1 arg1 :argument2 arg2}]
  (print arg1)
  (print arg2))
```

For years, I didn't know something like this was possible.

```
(defn test
  [&keys struct]
  (pp struct))
```

https://janet-lang.org/docs/functions.html should mention that what follows after `&keys` is actually a destructured struct.

sogaiu:
Thanks for elaborating.  Perhaps the text could expand on things a bit more.

Note though that currently the following text exists at the bottom of [the destructuring page](https://janet-lang.org/docs/destructuring.html):

> Destructuring works in many places in Janet, including let expressions, function parameters, and var.

amano-kenji:
I didn't recognize that there was struct destructuring after `&keys`....
I thought it was a special syntax...

CFiggers:
I thought the same thing! 

I saw your original issue over in janet-lang/janet (agree that this is a better place for it). Before that I didn't realize that `[&keys foo]` was valid syntax. 

This is a useful thing to know, because sometimes I need to pass the whole struct received after `&keys` to another function. But if I've destructured it in the main function's params, then there's no symbol bound to the whole struct and I have to manually reconstruct it in the arguments of the sub function:

```janet
(defn foo [&keys {:a apples :b bananas}]
  (print apples)
  (print bananas)
  (count-calories {:a apples :b bananas}))
```

This small example is already annoying, but it's much worse the more keys are in the struct.

That kinda turned me off from using `&keys` at all almost ever:

```janet
(defn foo [fruitmap]
  (print (fruitmap apples))
  (print (fruitmap bananas))
  (count-calories fruitmap))
```

But it turns out, this is valid:

```janet
(defn foo [&keys fruitmap]
  (print (fruitmap apples))
  (print (fruitmap bananas))
  (count-calories fruitmap))
```

So `&keys` might get more use in my Janet code moving forward.

What would be really nice is if I could have the best of both worlds:

```janet
(defn foo [&keys {:a apples :b bananas} :as fruitmap]
  (print apples)
  (print bananas)
  (count-calories fruitmap))
```

[...]

PR #214, opened 2024-02-04
Explain use of symbol after &keys (#205)

sogaiu:
This PR is an attempt to address #205.

It adds some text mentioning that a symbol can be placed after `&keys` (instead of a struct literal) and a code example that demonstrates the use.


pepe (approved):
LGTM

PR #224, opened 2024-06-18
Update look peg special docs

sogaiu:
This PR contains changes to the docs for the `look` peg special.

The docs do not currently appear to cover the case of `look` being called with only a single argument:

>    (look offset patt)

>    Matches only if patt matches at a fixed offset. offset can be any integer. patt will not produce captures and the peg will not advance any characters.

This PR attempts to cover this additional case.

Please see [here](https://github.com/janet-lang/janet/issues/1465) for some background.


sogaiu:
Thanks to @pepe for discussion :+1:

pepe (approved):
LGTM

PR #233, opened 2024-09-13
Add documentation on nth and only-tags PEG specials

katomuso:


katomuso:
> I think `nth` has an optional tag argument...at least that's what it looks like to me [here](https://github.com/janet-lang/janet/blob/88e60c309c0e05970cfed012f5b1eae98ad92bbb/src/core/peg.c#L502).
> 
> If that's the case, may be it's good to have that in the signature using `?tag` or some such? FWIW, right below the proposed last change is `uint` which has `?tag` at the end of its signature.

You are right, I've made the change.

Issue #223, opened 2024-06-16
Docs don't mention that function literals can be named with keywords


sogaiu:
Janet 1.35.0 gained the [ability to use keywords as names for anonymous functions](https://github.com/janet-lang/janet/commit/eb21d4fff48ee657ef83b7b4b54b386dd4f919fd).

May be at the very least we can add an example of this use in the section for `fn` on the [specials](https://janet-lang.org/docs/specials.html).

Possibly there could be explicity mention in the prose on that page or perhaps the [functions](https://janet-lang.org/docs/functions.html).

sogaiu:
Closing, as #235 has been merged.

PR #235, opened 2024-09-15
Document that fn can be named

sogaiu:
This PR is an attempt to address #223.

A couple of examples are added to the sample code for `fn` on the [Special forms page](https://janet-lang.org/1.36.0/docs/specials.html) showing that `fn` can be named via a symbol or a keyword.


pepe (approved):
LGMT

Issue #232, opened 2024-09-13
Document fiber user signals


pnelson:
`fiber/status` may return `:user(0-7)` to indicate that the fiber is suspended by a user signal.

`defer` will also run `form` if a user signal 0-4 is received, I assume within the fiber it is being run in.

The `signal` function appears to be used to send user signals. The function documentation should probably make mention of "user" but I think the [example provided on Zulip](https://janet.zulipchat.com/#narrow/stream/409517-help/topic/How.20to.20suspend.20a.20fiber.20with.20a.20user.20signal.3F/near/469823571) would make a great addition to the Fiber Overview page:

```janet
(defer (pp :hello) (signal :user1 :smile) (pp :oops))
:hello
user1: smile
  in signal [src/core/corelib.c] on line 701
  in defer [repl] on line 2, column 20
  in thunk [repl] (tail call) on line 2, column 1
```

sogaiu:
I've started going over some of the related website docs and docstrings.

It does seem sensible to me to add mention of `signal` to the [current version of the Fiber Overview page](https://janet-lang.org/1.36.0/docs/fibers/index.html) as it doesn't appear to be there currently, possibly with some example code as you suggested.

Regarding the function docs for `signal`, [its current definition](https://github.com/janet-lang/janet/blob/88e60c309c0e05970cfed012f5b1eae98ad92bbb/src/core/corelib.c#L701-L721) is:

```c

JANET_CORE_FN(janet_core_signal,
              "(signal what x)",
              "Raise a signal with payload x. ") {
    janet_arity(argc, 1, 2);
    Janet payload = argc == 2 ? argv[1] : janet_wrap_nil();
    if (janet_checkint(argv[0])) {
        int32_t s = janet_unwrap_integer(argv[0]);
        if (s < 0 || s > 9) {
            janet_panicf("expected user signal between 0 and 9, got %d", s);
        }
        janet_signalv(JANET_SIGNAL_USER0 + s, payload);
    } else {
        JanetKeyword kw = janet_getkeyword(argv, 0);
        for (unsigned i = 0; i < sizeof(janet_signal_names) / sizeof(char *); i++) {
            if (!janet_cstrcmp(kw, janet_signal_names[i])) {
                janet_signalv((JanetSignal) i, payload);
            }
        }
    }
    janet_panicf("unknown signal %v", argv[0]);
}
```

My impression of that code is that the intention is for one to pass an integer or a keyword to `signal` for the value of `what`.

[...]

PR #236, opened 2024-09-16
Mention signal in fiber overview

sogaiu:
This PR is an attempt to address #232.

My current hope is to pair it with an update to `signal`'s docstring with something like what's described [here](https://github.com/janet-lang/janet-lang.org/issues/232#issuecomment-2351351905).


pepe (approved):
LGTM

PR #240, opened 2024-12-03
Add dark theme

oliverfencott:
Hello, during my quest to learn Janet I often find myself browsing the site late at night and it's quite a strain on my eyes.
This PR minimally changes the 3 primary CSS files by adding media queries that use a dark theme when the user's OS/browser is set to dark.

In keeping with the style of the codebase, I've opted to inline colour values as opposed to using CSS variables. The majority of colours I've chosen are already present in the styling for the regular theme, save for one or two hand-picked shades of grey where necessary.

I've chosen to co-locate the media queries next to the classnames/ids that are affected. This leads to duplicated code (the same media selector) but simpler navigation.

Attached are some (very large) screenshots to give an idea of how the site looks with the changes.

homepage

docs

api

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -301,9 +301,7 @@ code {
 
 @media (prefers-color-scheme: dark) {
     .mendoza-code {
-        padding: 0 5px;
         background: #333;
-        white-space: nowrap;
     }
 }
 
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -16,31 +16,36 @@ article, aside, canvas, details, embed,
 figure, figcaption, footer, header, hgroup,
 menu, nav, output, ruby, section, summary,
 time, mark, audio, video {
-	margin: 0;
-	padding: 0;
-	border: 0;
-	font-size: 100%;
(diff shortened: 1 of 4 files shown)


static/css/main.css
```diff
@@ -254,10 +295,18 @@ code {
...
+    background: #eee;
     white-space: nowrap;
 }
 
+@media (prefers-color-scheme: dark) {
+    .mendoza-code {
+        padding: 0 5px;
+        background: #333;
+        white-space: nowrap;
+    }
```
oliverfencott, line 307: Padding and whitespace are duplicated.

sogaiu:
FWIW, as a work-around, I've been using https://darkreader.org/ with pretty good results.

(Not opposed to dark theme support, just mentioning :) )

oliverfencott:
> FWIW, as a work-around, I've been using https://darkreader.org/ with pretty good results.
> 
> (Not opposed to dark theme support, just mentioning :) )

Thank you for the recommendation, I'll be sure to check it out.

PR #242, opened 2025-01-02
Adding examples to Janet api documentation.

erichaney:
Added an example for `++`, `--`, and `->`. Let me know if the filenames are not what they ought to be.

### DIFF
--- a/examples/++.janet
+++ b/examples/++.janet
@@ -1,0 +1,3 @@
+(var a 0) # -> 0
+(set a (+ a 1)) # -> 1
+(++ a) # -> 2 (same effect as above)
--- a/examples/--.janet
+++ b/examples/--.janet
@@ -1,0 +1,3 @@
+(var x 10) # -> 10
+(set x (- x 1)) # -> 9
+(-- x) # -> 8 (same effect as above)
--- a/examples/-_62.janet
+++ b/examples/-_62.janet
@@ -1,0 +1,5 @@
+(-> {:a [1 2 3] :b [4 5 6]}
+    (get :a)
+    (sum)
+    ((fn [x y] (/ x y)) 2)) # The 6 from the sum is threaded first and binds to x, then 2 binds to y.
+# -> 3


bakpakin:
Thanks Eric, I appreciate the code examples!

Looks like the example file for `->` should just be `examples/->.janet`, no need for the escape. To check, run `make; make run` and verify that examples pop up in the generated docs.

erichaney:
I hesitated to name the file `->.janet` since > and < are not permitted in [file names for windows](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file).

But I can see that what I did isn't getting picked up in the generated docs. No biggie. Feel free to close this and I can resubmit with the unescaped file name.

CFiggers:
@erichaney You can just change it in place and the PR will update automatically. 🙂

sogaiu:
> I hesitated to name the file ->.janet since > and < are not permitted in [file names for windows](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file).

I also thought this when looking at this before.

Some of the characters that looked potentially problematic to me included:

* `/`
* `<`
* `>`
* `*`
* `%`
* `:`
* `?`

FWIW, I think some of the code that handles symbol to file name conversion is [here](https://github.com/janet-lang/janet-lang.org/blob/fa0ae926c461d3c7fbf0026a4410212ffa12f74c/content/api/gen-docs.janet#L14-L18).

My translation of that is:

```janet
(def- replacer
  (peg/compile
    ~(accumulate (any (choice (replace (capture (set "/*%"))
                                       ,|(string "_" (0 $))) 
                              (capture 1))))))

(defn- sym-to-filename
  "Convert a symbol to a filename. Certain filenames are not allowed on various operating systems."
  [fname]
  (string "examples/" ((peg/match replacer fname) 0) ".janet"))
```

For comparison, [this](https://github.com/sogaiu/janet-ref/blob/942be1fee5b1fd997f0093a3638c5db05bd50d55/janet-ref/things.janet#L3-L21) is the scheme I settled on for [janet-ref](https://github.com/sogaiu/janet-ref).

erichaney:
I removed the escape from the file name. I'll focus on contributing examples that don't use any special characters for now. 👍

sogaiu:
I think the evaluation result for [this line](https://github.com/janet-lang/janet-lang.org/pull/242/files#diff-993d1e97ab738d791ed832cadf3eef4d759b49d126a029cd74a2956e42e11a0aR4) (line 4 of `-.janet`) is off:

```janet
(- 1.4 -4.5) # -> -5.9
```

What I get here is:

```
$ janet
repl:1:> (- 1.4 -4.5)
5.9
```

sogaiu:
For [this line](https://github.com/janet-lang/janet-lang.org/pull/242/files#diff-03a9edbef6da5460afe1b6e578f983f9db3b70e3c788926c6840a3fdd5882530R4) (line 4 of `accumulate2.janet`):

```janet
(reduce2 + []) # -> 0
```

what I get here is:

```
$ janet
repl:1:> (reduce2 + [])
nil
```

sogaiu:
Not sure how much we should be concerned with the following sort of thing, but FWIW, I noticed in some lines that there is no space between the `#` and the `->` (i.e. I see `#->` instead of `# ->`).  For the most part, I think elsewhere in the examples (also outside of this PR) there has been use of a space.

For reference, the following have examples of the aforementioned:

* `accumulate.janet` (line 1)
* `accumulate2.janet` (line 1)
* `all.janet` (lines 1 and 2)

Issue #244, opened 2025-01-11
Naming scheme for examples files non-friendly to Windows


sogaiu:
As mentioned in [this comment](https://github.com/janet-lang/janet-lang.org/pull/242#issuecomment-2568621895), some characters don't really work well as parts of filenames in typical Windows setups.  One consequence of this might be that folks who are using Windows may be at a disadvantage working with this repository (including participants in this year's Janetuary examples activity).

[This comment](https://github.com/janet-lang/janet-lang.org/pull/242#issuecomment-2568631883) attempts to enumerate specifically which characters might be problematic.

So far `/`, `*`, and `%` receive special handling, but:

* `<`
* `>`
* `:`
* `?`

don't seem to.

May be tweaking the code [here](https://github.com/janet-lang/janet-lang.org/blob/fa0ae926c461d3c7fbf0026a4410212ffa12f74c/content/api/gen-docs.janet#L14-L18) to handle these additional characters is an appropriate modification?

If the code is changed after some files with the additional characters in their names get created, the relevant filenames probably should get changed appropriately as well.

PR #247, opened 2025-01-11
Make examples file naming friendlier to Windows

sogaiu:
This PR is an attempt to address #244.

Please see that issue for some background.

---

If this PR is merged before https://github.com/janet-lang/janet-lang.org/pull/242, one of the files in that PR should probably have its filename changed.

The file in question is [this one](https://github.com/janet-lang/janet-lang.org/pull/242/files#diff-0c5d311e823c84570b14cdc6e4fb9444fac80719088c40b72d3cf090f7ecc1d5).  It is currently named `->.janet`, but likely a better name might end up being `-_62.janet`.

### DIFF
--- a/content/api/gen-docs.janet
+++ b/content/api/gen-docs.janet
@@ -11,7 +11,7 @@
     (set ret (string/replace "  " " " ret)))
   ret)
 
-(def- replacer (peg/compile ~(% (any (+ (/ '(set "/*%") ,|(string "_" (0 $))) '1)))))
+(def- replacer (peg/compile ~(% (any (+ (/ '(set "%*/:<>?") ,|(string "_" (0 $))) '1)))))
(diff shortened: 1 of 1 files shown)


bakpakin:
Yup, this looks fine to me. Since this is a Makefile based project I don't really use windows for this but that could be revisited.

sogaiu:
Re: operation on Windows

As a data point, I tried `make` via a msys2 mingw64 shell [1] (the built `janet.exe` reports this as `:cygwin` for some reason).

The result was that mongoose wouldn't build (presumably this is for circlet).

If I can figure out how to set up a `janet.exe` that detects things as `:mingw`, I might try that too.

---

[1] I think this was set up via `scoop`'s `mingw` package.

PR #254, opened 2025-01-14
Update README naming convention for examples

erichaney:


erichaney:
That's a good point.

I checked JanetDocs and it looks like you had written a good example of this function. If you are okay with it, I have included your example as part of this PR.

erichaney:
On second thought, I will close this PR and open an issue instead. It makes more sense for the maintainers to control how they would like the README to look.

PR #255, opened 2025-01-15
add examples for has-key? and has-value?

rwtolbert:
add examples for has-key? and has-value?


sogaiu:
Thanks!

I made one minor edit.

---

On a side note, if it's doable for you, please consider one file per PR.

rwtolbert:
@sogaiu Got it. Was being lazy since these are connected.

PR #258, opened 2025-01-16
add examples for keep

rwtolbert:
add examples for keep

### DIFF
--- a/examples/keep.janet
+++ b/examples/keep.janet
@@ -1,0 +1,8 @@
+# keep values > 1, equivalent to filter
+(keep (fn [x] (when (> x 1) x)) @[0 1 2 3]) # -> @[2 3]
+
+# for all members > 2, keep the square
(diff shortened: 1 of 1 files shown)


sogaiu:
@rwtolbert @bakpakin 

Although the last example sort of works:

```janet
# operates on values in table
(keep (fn [x] (when (> x 1) (* x x))) @{:foo 1 :bar 2 :baz 3}) # -> @[4 9]
```

I'm not sure it's such a good idea to advertise this.

Depending on the Janet version, the result may differ:

```janet
$ janet
repl:1:> (map + {:a 1 :b 2})
@[2 1]
```

```
$ janet
repl:1:> (map + {:a 1 :b 2})
@[2 1]
```

The docstring mentions "indexed collection `ind`" which arguably applies to "array and tuple" as `indexed?`'s docstring is:

```
Check if x is an array or tuple.
```

But should that be the case for a table or struct?

rwtolbert:
Yeah, I just added it because it "worked" but I can see the advantage of making it fail in the core and then remove this example.

PR #259, opened 2025-01-17
Swap an example for keep

sogaiu:
This PR swaps an example of `keep`.

The removal of one example is motivated by some comments starting [here](https://github.com/janet-lang/janet-lang.org/pull/258#issuecomment-2595925367).

Motivation for the additional example includes the points:

1. Use of a symbol in the predicate argument position that represents a pre-existing function instead of using a function constructed via the use of `fn` (as there are already examples of the latter)
2. Showing a use of `keep` which is in some sense about as simple as things might get

### DIFF
--- a/examples/keep.janet
+++ b/examples/keep.janet
@@ -1,8 +1,9 @@
+# returns an array of truthy results of predicate
+(keep identity [false :x nil true]) # -> @[:x true]
+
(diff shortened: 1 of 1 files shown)


sogaiu:
Thanks for taking a look!

PR #261, opened 2025-01-17
Add example for accumulate and accumulate2

erichaney:


sogaiu:
I do wonder about the behavior of `reduce2` in this example:

```janet
(reduce2 + []) # -> nil
```

I wonder if it could be considered a bug.

`reduce2`'s docstring is currently:

```
    The 2-argument version of `reduce` that does not take an 
    initialization value. Instead, the first element of the array is 
    used for initialization.
```

`[]` has no first element so the intent doesn't seem explicitly specified according to the docstring.

In these kinds of cases, it seems preferable to me to not record these as examples until further clarification.

sogaiu:
For comparison, `reduce` in Clojure can be called in two ways:

```
(reduce f coll)   (reduce f val coll)

f should be a function of 2 arguments. If val is not supplied,
returns the result of applying f to the first 2 items in coll, then
applying f to that result and the 3rd item, etc. If coll contains no
items, f must accept no arguments as well, and reduce returns the
result of calling f with no arguments.
```

The actual behavior in the following case (2-argument call) seems to match the docstring:

```
$ clj
Clojure 1.12.0
user=> (reduce + [])
0
```

sogaiu:
It looks like [the current behavior is intentional](https://github.com/janet-lang/janet/commit/65379741f70e2973aeda1d4954e8ce7e997a9816).

I think it might be better to change the docstring or the behavior.  Will make an issue.

sogaiu:
We received clarification in [this issue](https://github.com/janet-lang/janet/issues/1545) and the docstring has been updated to match the current behavior.

I've made one minor whitespace related edit.

---

For future readers, let's try to have one file per PR for the examples as mentioned in [these steps](https://github.com/sogaiu/janet-examples/blob/eb473c84f71a0389aa77453e404e2f026c6b1a9b/doc/janetuary-examples-2025.md#the-detailed-version).

I think it should make review / revision go more smoothly (e.g. if there is more than one file per PR and only one of them has an issue, it can hold up merging of the other content).

PR #264, opened 2025-01-18
Mention new scopes for if and while

sogaiu:
In a [Zulip discussion about the `if` special form](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/Scope.20within.20'if'.20special.20form), it came up that the [Special Forms page](https://janet-lang.org/1.37.1/docs/specials.html) doesn't appear to mention that `if` and `while` can introduce new lexical scopes (though the coverage for `do` and `fn` touch on the topic).

This PR suggests the addition of some text to remedy the situation.

### DIFF
--- a/content/docs/specials.mdz
+++ b/content/docs/specials.mdz
@@ -155,6 +155,9 @@ macro.
 The condition is considered false only if it evaluates to @code`nil` or
 @code`false` - all other values are considered @code`true`.
 
+If when-true or when-false is evaluated, this is done in the context
+of a newly introduced lexical scope.
+
 @codeblock[janet]```
 (if true 10) # evaluates to 10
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

rwtolbert:
I think `when` does the same as `if`.

sogaiu:
Yes, I think `when` does.

IIUC, that is because it's a macro that expands to `if`:

```
$ janet
repl:1:> (macex1 '(when true :hello))
(if true (do :hello))
```

However, this PR is specifically about the [current Special Forms page](https://janet-lang.org/1.37.1/docs/specials.html) and I think because `when` is a macro (and not a special form), it's not covered.

PR #267, opened 2025-01-23
add examples for drop-until

rwtolbert:
add examples for drop-until


sogaiu:
Similar to #268, pending some feedback on https://github.com/janet-lang/janet/issues/1548, let's make this a draft.

PR #268, opened 2025-01-23
add examples for drop-while

rwtolbert:
add examples for drop-while


sogaiu:
Pending some feedback on https://github.com/janet-lang/janet/issues/1548, let's make this a draft.

PR #269, opened 2025-01-23
add examples for filter

rwtolbert:
add examples for filter

### DIFF
--- a/examples/filter.janet
+++ b/examples/filter.janet
@@ -1,0 +1,9 @@
+(filter pos? [1 2 3 0 -4 5 6]) # -> @[1 2 3 5 6]
+
+(filter |(> (length $) 3) ["hello" "goodbye" "hi"]) # -> @["hello" "goodbye"]
+
+(filter |(< 65 $) "foo01bar") # -> @[102 111 111 98 97 114]
+
+(string/join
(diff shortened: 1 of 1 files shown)


sogaiu:
Looks good, though here are some alternates for consideration (note that `chr` happens at compile time):

```janet
(filter |(< 65 $) "foo01bar") 
# => 
@[102 111 111 98 97 114]

(filter |(< (chr "A") $) "foo01bar") 
# => 
@[102 111 111 98 97 114]
```

May be using `chr` is somewhat clearer and if folks don't know what that is, searching for it on the Core API page or via `doc` wouldn't be too bad?

```janet
(string/join
  (map string/from-bytes 
       (filter |(< "A" (string/from-bytes $)) "foo01bar"))) 
# =>
"foobar"

(string/from-bytes ;(filter |(< (chr "A") $) "foo01bar"))
# =>
"foobar"
```

Perhaps this way we can avoid `string/from-bytes` being called more than once and avoid a few other calls?

rwtolbert:
I like that idea. Let me see what I can do.

rwtolbert:
@sogaiu updated.

sogaiu:
Thanks!

PR #270, opened 2025-01-23
add examples for delay

rwtolbert:
`delay` doesn't seem to map to the simple example format of `(foo) # -> output` Also, we don't really show `repl>` in other examples (maybe we should?)

Certainly open for suggestions.


rwtolbert:
Pushed an update to mimic style from `defer`

sogaiu:
Thanks!

PR #271, opened 2025-01-24
Tweak % examples

sogaiu:
This PR suggests tweaking some of the examples for `%`.

Looking at the first four, I got the sense that expressing all combinations of signs for the arguments was the intent:

```janet
(% 10 3) # -> 1
(% -10 3) # -> -1
(% 10 -3) # -> 1
(% -10 -3) # -> -1
```

The second set of four examples has been tweaked to match the pattern (as otherwise there are duplicate examples).

A similar change in sign was applied to `(% -1.4 0)` to make it `(% 1.4 0)` so that there were two distinct examples.

The last two examples also use `nan?` to make the result possibly less confusing.


pepe (approved):
LGTM

PR #278, opened 2025-01-27
Fix table background in dark mode

elfenermarcell:
Before 
image
After
image

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -207,6 +207,12 @@ table tr {
     background-color: white;
 }
 
+@media (prefers-color-scheme: dark) {
+    table tr {
+        background: #222;
+    }
+}
+
 table tr th {
     font-weight: bold;
     border: 1px solid #cccccc;


sogaiu:
I think the intended effect is an improvement over the current arrangement which I believe was brought into place in #240.

When I tested the changes in [d522cdf](https://github.com/janet-lang/janet-lang.org/pull/278/commits/d522cdfdb0d58865024b0ffb03c9ea6553629cba) locally, I got the sense that they don't work for older versions of the docs though [1], e.g. http://localhost:8000/1.36.0/docs/peg.html:

dark-mode-testing

Perhaps it is the case that #240 also didn't account for older versions?

@oliverfencott How does it look to you?

oliverfencott:
> Perhaps it is the case that #240 also didn't account for older versions?
> 
> @oliverfencott How does it look to you?

Looks good to me!

You're correct that dark mode doesn't work for older versions. I've started a branch that will fix this, though; I'll open a PR later this week for it.

elfenermarcell:
See https://github.com/janet-lang/janet-lang.org/pull/279

PR #230, opened 2024-09-06
Proposal: Restructure Janet's Documentation

CFiggers:
This PR proposes restructuring janet-lang.org's documentation by creating two (2) new top-level sections, one (1) for jpm and the other for Spork.

## Rationale

The recent decisions to include function references for Spork and jpm on janet-lang.org make sense to me, both from an ease-of-maintenance and ease-of-discovery standpoint. However, the current integration is not ideal. 

Regarding jpm: The main body of documentation for jpm is in [a single (decently long) page](https://janet-lang.org/docs/jpm.html) under the Documentation section. Separate from that page, there's also the [recently-included grouping of jpm API references](https://janet-lang.org/api/jpm/index.html) under the API section. In my opinion, the jpm-related pages—explanatory docs and api references—should live close together rather than being split up. In fact, on reflection, the page about jpm under the Documentation section is a bit of an oddball compared with the rest of the pages in the section. The other pages all have to do with the syntax, semantics, and key concepts of _the language itself_. Jpm, while important in practice, is not technically speaking a part of the language itself (and is even decreasing in critical importance with the introduction of the `bundle/*` functions to the core library). And on top of that, because of the large amount of ground that one page has to cover, that page is on the longer end of pages on the current site—it could easily (indeed, I feel it should) be split into multiple pages for easier indexing in order to make its contents more visible in the overall site's ToC.

Regarding Spork: Currently, because there are more individual modules for Spork than there are individual modules in the core library, the Spork docs visually overpower the core API's module pages. On my laptop at the default zoom level, Spork module pages alone take up an entire screen from top to bottom, meaning the first couple of core modules are distantly separated from the last few (by more than a screen height, in my specific case, so it's not even possible to visually inspect the full list of core modules without scrolling). This is not beginner-friendly. Most beginners need to familiarize themselves with the language's core functions and macros (i.e. answering the question "What does this language have in it, anyway?") before venturing into the "expanded core lib" that Spork is shaping into. (The main API page containing all functions is still perfectly accessible, of course, but speaking from my own experience the list of module pages is still a helpful tool for discovery of the core library's contents.)

## Summary of Proposal

To fix both sets of issues outlined above, I propose that we:

* Create new top-level sections for jpm and Spork
* Move the documentation on jpm that currently lives under the "Documentation" section into the new JPM section and also split it into multiple pages for easier reference (there are new sections on that page that I didn't realize had been added until I was going through this exercise!)
* Move the API references for jpm and Spork into their respective newly-created sections
[...]

PR #280, opened 2025-01-31
Add example for spork/argparse/argparse

brendanhowell:
I am a janet noob so please let me know if I should change anything style-wise.

### DIFF
--- a/examples/spork_47argparse_47argparse.janet
+++ b/examples/spork_47argparse_47argparse.janet
@@ -1,0 +1,23 @@
+(import spork)
+
+(def argparse-params
+  ["An example CLI tool."
+   "my-flag" {:kind :flag
+              :short "f"
+              :help "set my-flag"}
+   "my-opt" {:kind :option
+             :short "o"
+             :help "Value for option"
+             :default "DEFAULT VALUE"}
+   "my-req" {:kind :option
+             :short "r"
+             :required true}])
+
+(def cli-args (spork/argparse/argparse ;argparse-params))
+
+(unless cli-args
+  (os/exit 1))
+
+(print "flag: " (get cli-args "my-flag"))
+(print "my-opt: " (get cli-args "my-opt"))
+(print "my-req: " (get cli-args "my-req"))


sogaiu:
Thanks for your contribution.

I've tried the code and it seems to work :+1: 

Below are some general concerns that this PR has surfaced -- I don't think they are points about the content of this PR.

---

IIUC, this is the first instance of examples for spork and I'm not sure how we want to handle them.  Some concerns might be:

* Do we really want spork examples to live in the same directory as those for janet itself?
* Do we want the examples to live in the janet-lang.org repository?  May be the answer is "yes".
* [This PR](https://github.com/janet-lang/janet-lang.org/pull/230) may result in some changes to how things are arranged.  Are there any factors we should consider related to that?

---

One thing I noticed is that although the examples show up on the overall spork page at:

```
http://localhost:8000/api/spork/argparse.html
```

they don't seem to show up on the page for `spork/argparse` at the appropriate spot:

```
http://localhost:8000/api/spork/index.html#spork/argparse/argparse
```

Note that a similar thing does not appear to hold for some janet core items.  For example, [`string/split` appears on both the overall page for core items](https://janet-lang.org/api/index.html#string/split) as well as [on the String Module page](https://janet-lang.org/api/string.html#string/split).

brendanhowell:
> IIUC, this is the first instance of examples for spork and I'm not sure how we want to handle them. Some concerns might be:
> 
>     * Do we really want spork examples to live in the same directory as those for janet itself?
> 
>     * Do we want the examples to live in the janet-lang.org repository?  May be the answer is "yes".
> 
>     * [This PR](https://github.com/janet-lang/janet-lang.org/pull/230) may result in some changes to how things are arranged.  Are there any factors we should consider related to that?
> 

Hmm… I was wondering some of the same things. The proposed PR sounds quite sensible to me. But it looks like it needs a little work. I'm happy to move my example to fit the new organization. Maybe we can leave this open and I will adjust once things get merged.

> One thing I noticed is that although the examples show up on the overall spork page at:
> 
> ```
> http://localhost:8000/api/spork/argparse.html
> ```
> 
> they don't seem to show up on the page for `spork/argparse` at the appropriate spot:
> 
> ```
> http://localhost:8000/api/spork/index.html#spork/argparse/argparse
> ```
> 
> Note that a similar thing does not appear to hold for some janet core items. For example, `string/split` appears on both [the overall page for core items](https://janet-lang.org/api/index.html#string/split) as well as [on the String Module page](https://janet-lang.org/api/string.html#string/split).

I now see that I actually have the opposite problem but maybe that's what you meant. I can see my example on the overall spork page but not on the sub-page for argparse. Should I file a bug or just wait to see how things look after the re-organization?

sogaiu:
> Maybe we can leave this open and I will adjust once things get merged.

Why don't we go with this for the moment.

> I now see that I actually have the opposite problem but maybe that's what you meant. I can see my example on the overall spork page but not on the sub-page for argparse. 

Sorry I got the URLs mixed up (^^;

I haved edited my post above to fix it.

> Should I file a bug or just wait to see how things look after the re-organization?

If you don't mind doing it please file an issue.  I think having an issue for it separately might make it easier to track as well as adjust for if necessary since it's unknown when the re-organization will take place.

brendanhowell:
Relevant bug report #282

PR #284, opened 2025-02-04
add examples for each

rwtolbert:
add examples for each


sogaiu:
The [current `loop` examples use the phrase "in an unspecified order"](https://janet-lang.org/1.37.1/api/index.html#loop), may be that could work here too?

```janet
# prints values 1 and 2 in an unspecified order
(each x {:a 1 :b 2} (prin x)) # -> nil
```

sogaiu:
Looks good!

I've made one minor edit.

rwtolbert:
👍

PR #285, opened 2025-02-04
add examples for eachk

rwtolbert:
add examples for eachk


sogaiu:
May be using `[:a :b :c :d :e]` as below instead of `[1 2 3 4 5]` still provides enough "surprise" and is slightly less confusing?

```janet
# prints 0 1 2 3 4
(eachk i [:a :b :c :d :e] (prin i " ")) # -> nil
```

PR #287, opened 2025-02-05
Improve consistency of examples

sogaiu:
This PR is an attempt to improve the consistency of some of the existing examples.

Apart from a minor adjustment to correct an example, the changes basically are of three types:

1. Try to use the `# -> <value>` comment to describe return values only
2. Try to use the `# -> error: ...` (or `# -> compile error: ...`, etc.) comment(s) to describe errors
3. Use comments to describe printed output -- AFAICT, this is more in line with how most of the other examples do things

### DIFF
--- a/examples/defn-.janet
+++ b/examples/defn-.janet
@@ -7,7 +7,7 @@
 # In a file main.janet
 (import module)
 
-(module/not-exposed-fn 10) # -> Unknown symbol error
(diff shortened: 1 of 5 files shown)


pepe (approved):
LGTM

rwtolbert:
+1

PR #288, opened 2025-02-05
Tweak examples: *, apply, and product

sogaiu:
This PR contains some minor tweaks to examples for `*`, `apply`, and `product`.

Apart from some formatting, some return values and an error message are modified.

### DIFF
--- a/examples/_42.janet
+++ b/examples/_42.janet
@@ -1,8 +1,8 @@
 (*) # -> 1
 (* 10) # -> 10
 (* 10 20) # -> 200
-(* 1 2 3 4 5 6 7) #-> 5040
+(* 1 2 3 4 5 6 7) # -> 5040
 
 # Can take product of array with splice, but 'product' is better
-(* ;(range 1 20)) #-> 1.21645e+17
-(product (range 1 20)) #-> 1.21645e+17
+(* ;(range 1 20)) # -> 1.21645100408832e+17
+(product (range 1 20)) # -> 1.21645100408832e+17
--- a/examples/apply.janet
+++ b/examples/apply.janet
@@ -1,7 +1,7 @@
 (apply + (range 10)) # -> 45
 (apply + []) # -> 0
 (apply + 1 2 3 4 5 6 7 [8 9 10]) # -> 55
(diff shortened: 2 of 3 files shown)


rwtolbert:
@sogaiu I've got mixed feelings about `product` applied to a table. Just because it works, it feels like an unplanned consequence not a real feature.

Seem if a user really wants to `sum` or `product` the values from a table, then forcing this syntax makes it more clear:

```janet
repl:11:> (def foo @{:a 2 :b 3})
@{:a 2 :b 3}
repl:12:> (product (values foo))
6
```

Just my 2c

sogaiu:
@rwtolbert I feel similarly (^^;

FWIW, I did some digging and learned that the original example was added in [this commit](https://github.com/janet-lang/janet-lang.org/commit/ae6d128e71947a85b026e27e141f352eb426f6dc#diff-5956b7af633c78433a6b5326d7374b5a1125f2a26a0b678e829a59b621200a03) (along with a similar example for `sum`).  Given the author of that commit, I suspect that the current behavior is intentional.

rwtolbert:
In some ways it is good that at least we document this behavior.

the corner case that concerns me is some function that depends on the order of the input that when applied like this then fails in odd ways with the non-deterministic order of the values. But then `(values foo)` doesn't promise any order either.

sogaiu:
With respect to `product` and `sum`, the underlying functions being applied are `*` and `+`, and ordinarily [1] IIUC the order of application across a set of elements shouldn't matter for these(?).

AFAIU, [`product`](https://github.com/janet-lang/janet/blob/410f8d69bc603558de987a0a4ba7877281f2a287/src/boot/boot.janet#L738-L743) and [`sum`](https://github.com/janet-lang/janet/blob/410f8d69bc603558de987a0a4ba7877281f2a287/src/boot/boot.janet#L708-L713) both currently use `each` in their implementations, and `each` in turn uses `next` along with a succession of keys to traverse various data structures (which include indexed and dictionary types).  If we think of indexed and dictionary types as being (typically) different only in the kinds of keys they have [2], may be it's not so strange for `sum` and `product` to be able to handle both?

---

[1] Not thinking about floating point related errors.

[2] That indexed things are "sorted" being ignored here while dictionaries (at least in Janet) are not.

PR #289, opened 2025-03-06
Prevent recent text addition from being interpreted as a Janet statement

engenforge:
Previous commit [a334121](https://github.com/janet-lang/janet-lang.org/commit/a3341215b5a4515169df923cc2fe69de0da969b1) breaks the build on my system (and I assume for everyone else).

It appears that Mendoza is interpreting the "(you will need ..." text as a Janet statement.  This change prevents that and enables the build to complete.


sogaiu:
I tried the suggested change.

The build succeeded here and the resulting page seemed fine when view in a web browser.

PR #293, opened 2025-04-09
Clarify meaning of $&

sogaiu:
This PR is a suggestion to clarify the meaning of `$&` as [currently expressed on the function page](https://janet-lang.org/1.38.0/docs/functions.html):

```
Similarly, $0, $1, etc. refer to the arguments at index 0, 1,
etc. and $& refers to all passed arguments as a tuple.
```

IIUC, when `$`, `$0`, ..., `$i` is/are used, `$&` refers to less than "all passed arguments":

```janet
(|[$0 $1 $&] 1 0 -1)
# =>
[1 0 [-1]]

(|[$ $&] 1 0)
# =>
[1 [0]]
```

The suggested change is to express the above as:

```
Similarly, $0, $1, etc. refer to the arguments at index 0, 1,
etc. and $& refers to all remaining arguments as a tuple.
```

i.e. saying "all remaining arguments" instead of "all passed arguments".


pepe (approved):
LGTM

PR #294, opened 2025-04-09
Add examples for count

sogaiu:
Here are some examples for `count`, including the use of multiple input data structures and an illustration of an edge case.


pepe (approved):
LGTM

PR #297, opened 2025-04-13
Add more examples for all

sogaiu:
Here are a couple more examples for `all` which illustrate use of multiple input data structures.


pepe (approved):
LGTM

PR #304, opened 2025-08-02
Change janetdocs.com to .org

veqqq:
I hope these two templates cover it.

I wasn't sure whether I should change all the saved html - I was surprised by its presence in the repo instead of only templates. Swlkr said he will redirect .com to .org, so eventually the old ones will break (due to my new url scheme with package names) (deprecated bindings will break on the change, though). If you decide to change them, ctrl-f:

-  `janetdocs.com/` -> `janetdocs.org/core-api/` (note the `/`)
- `janetdocs.com"` -> `janetdocs.org"` (for indexes, note the `"`)

### DIFF
--- a/content/api/gen-docs.janet
+++ b/content/api/gen-docs.janet
@@ -24,7 +24,8 @@
 (def- url-repl-chars
   {(chr "%") "%25"
    (chr "?") "_q"
-   (chr "=") "%3d"})
+   (chr "=") "%3d"
+   (chr "/") "%2f"})
 
 (defn jdoc-escape
   [str]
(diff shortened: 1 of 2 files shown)


sogaiu:
> I wasn't sure whether I should change all the saved html - I was surprised by its presence in the repo instead of only templates.

I think it's probably worth examining what is involved in modifying the saved `.html` files.

For reference, the following might give an idea for just 1.10.1:

```
$ rg -c Community\ Examples\</a 
api/debug.html:9
api/peg.html:2
api/parser.html:13
api/net.html:6
api/math.html:40
api/fiber.html:9
api/module.html:4
api/table.html:6
api/os.html:32
api/file.html:8
api/array.html:12
api/tuple.html:5
api/string.html:20
api/tarray.html:7
api/int.html:2
api/buffer.html:16
api/thread.html:5
api/index.html:461
doc.html:461
```

IIUC, 1.10.1 is the first version from which there were links to janetdocs.com.  I don't see such links for 1.9.1 or 1.8.1 and presume that means nothing lower has such links.

PR #314, opened 2026-01-19
Add a section about lazy evaluation.

amano-kenji:


sogaiu:
AFAIU, the current main prose of the website does not pick out specific portions of spork so some of the proposed changes here seem a little odd / mismatched stylistically.

I've thought of a couple of alternative approaches though:

1. There is already [a section on fibers](https://janet-lang.org/docs/fibers/index.html) and this would be one natural place to mention the existence of `spork/generators` along with a link to the docs perhaps.  Then details can be contained in the `spork/generators` docs.

2. Alternatively, there could be a page on eager evaluation to make explicit that that's really what the main semantics of Janet provide (i.e. it's not like Haskell which is basically the reverse).  Along with that text, `spork/generators` could be mentioned together with lazy evaluation and with a link to `spork/generators` docs for an example of how it might be done.  Then as above, details can be contained in the `spork/generators` docs.

I think either of those approaches aligns better with how the documentation is currently.

PR #326, opened 2026-02-07
Document spork/stream's standard stream functions

amano-kenji:
It closes https://github.com/janet-lang/janet/issues/1398


amano-kenji:
I mentioned windows.

PR #71, opened 2020-06-15
Additions and tweaks to the peg docs

sogaiu:
Attempted to document `repeat`, its alias, and `backmatch`.

Also includes some minor tweaks.

PR #336, opened 2026-04-26
docs/peg: improve documentation for `to` and `thru`.

### LINKED ISSUE #1742: There is no intuitive way to mimick `string/has-suffix?` with `peg/match`.
With regex, I can just write `.jpg$` to match a file name extension. With `peg/match`, I can't figure a way out.

peg is supposed to be better than regex, yet it makes some things nearly impossible.

Can peg be improved in this area?

amano-kenji:
It fixes https://github.com/janet-lang/janet/issues/1742

### DIFF
--- a/content/docs/peg.mdz
+++ b/content/docs/peg.mdz
@@ -187,13 +187,13 @@ compiled to bytecode).
         @td{ Alias for @code`(look offset ?patt)` }}
 
     @tr{@td{@code`(to patt)` }
-        @td{ Match up to patt (but not including it). If the end of the input
-             is reached and patt is not matched, the entire pattern does not
-             match. }}
+        @td{ Match until patt is found (not including it in the match). If the
+             end of the input is reached without finding patt, the entire
+             pattern does not match. }}
 
     @tr{@td{@code`(thru patt)` }
-        @td{ Match up through patt (thus including it). If the end of the
-             input is reached and patt is not matched, the entire pattern
+        @td{ Match until patt is found (including it in the match). If the end
+             of the input is reached without finding patt, the entire pattern
              does not match. }}
 
     @tr{@td{@code`(backmatch ?tag)` }


sogaiu:
AFAIU, the word "until" can mean inclusive or exclusive depending on the situation [1].  

The original explicitly says:

* `(but not including it)` for `to`
* `(thus including it)` for `thru`

That part of the proposed change actually makes this part of the text less clear.

I would prefer that those removed bits remain in some form.

Really, I find the original to be clearer, TBH.

A further complication is that there is also a PEG special called `til`.  I think if we could find a word other than "until", it might be less confusing.

---

[1] See [here](https://ell.stackexchange.com/a/33384) for example:

>  It's exclusive if the situation described is notable by its absence. It's likely to be inclusive if the situation described is notable by its presence.

amano-kenji:
I don't see `til`.

sogaiu:
@ianthehenry submitted a [PR for `til`](https://github.com/janet-lang/janet/pull/1528) and it was merged in February of 2025.

sogaiu:
I think the use of a single word "until" is not an improvement over the use of "match up to" and "match up through".  May be we can figure out a way to use different words for each case.

Using "to" in the documentation for `to` and "through" in the documentation of `thru` makes a lot of sense to me though.

Do the words "inclusive" and "exclusive" seem clearer?

amano-kenji:
What about this?

sogaiu:
This clearly deviates from the style of the rest of the documentation.  I don't think this is an improvement.

Did you see the idea of "inclusive" vs "exclusive"?  Are those words clear to you?

amano-kenji:
I prefer plain verbose procedural english even if it looks clumsy.

Style doesn't guarantee clarity.

Is my explanation clearer?

sogaiu:
It is overly verbose compared with the rest of the documentation on the page.

Your preferences alone should not dictate the changes.

amano-kenji:
Do you have a better idea?

sogaiu:
You have not answered my question about "inclusive" vs "exclusive".

Please answer it.

amano-kenji:
I don't know what I'm supposed to think of inclusive and exclusive. English is my second language, and something stylish, concise, and clear is beyond my current capability. I'm not a trained technical writer. I'm just a dude.

amano-kenji:
I'm confused on how to improve this. I will need more time to think about it. If you have a better idea, feel free to suggest it.

By the way, the current approach is mathemtical induction. I was instructed to use mathematical induction in university math classes. Mathematical induction is easy on my psyche.

sogaiu:
My understanding is that in the original issue [you parsed the sentences in such a way as to group "match" with "up" (so "match up") without going further](https://github.com/janet-lang/janet/issues/1742#issuecomment-4318186569):

> What does "match up" mean exactly?

I think the intended way of parsing the original text is to treat "match up to" and "match up through" as units, not "match up" and then followed by "to" or "through".

Searching turns up usage of these phrases in the context of regular expressions:

"match up to":

* https://stackoverflow.com/questions/2013124/matching-up-to-the-first-occurrence-of-a-character-with-a-regular-expression
* https://vi.stackexchange.com/questions/19128/using-neovims-regex-to-match-up-to-but-excluding-a-certain-character

"match up through":

* https://stackoverflow.com/questions/32569032/sed-how-to-print-all-lines-after-line-with-a-match
* https://stackoverflow.com/questions/19940935/extract-email-addresses-from-text-file-using-regex-with-bash-or-command-line

I think it's not unreasonable to interpret that in the context of text matching these phrases are not confusing to at least some developers who speak English.  I think you might be better off learning about them because they seem to be used elsewhere too.  So one idea is to keep the documentation as-is.

Another idea is to drop the word "up" and use "until" in one of the cases (and still retain the parenthetical comments).  That might have prevented your original parsing from happening.

The changed text would look like this:

> Match until patt (but not including it). If the end of the input is reached and patt is not matched, the entire pattern does not match.

and:

> Match through patt (thus including it). If the end of the input is reached and patt is not matched, the entire pattern does not match.

I find these to still be comprehensible and stylistically consistent with the rest of the documentation.

amano-kenji:
`up to` is a known idiom, so I can understand it. But, `up through` is not a known english idiom. I was confused by this.

I don't like idioms.. I want clarity. If I was a language designer, I would delete the notion of idioms.

amano-kenji:
What about this?

* to: Match until patt (but not including it).
* thru: Match until patt. Then, also match patt.

If I didn't know anything, `match through` would be confusing because it doesn't make me imagine `.*patt`.

`.*patt` wouldn't be a natural interpretation of `match through patt`.

It's said programs written in scala language are difficult to understand because scala programmers focus on conciseness without considering readability. Conciseness shouldn't be prioritized above clarity.

It actually takes less human time to read more characters that are clearer than to read concise riddles that confuse people. A riddler is a bad guy. Scala programmers are riddlers.

amano-kenji:
I found others ways to phrase them.

* to
  * Match characters while patt doesn't match.
  * Match until patt(not including it).
  * Match individual characters while patt doesn't match.
  * Match individual characters until patt(not included in the match) matches.
* thru
  * Match characters until patt is found and included in the match.
  * Match individual characters until patt is found and included in the match.

amano-kenji:
I think I have optimized both conciseness and clarity. Is it good enough now?

sogaiu:
If one were to go for the following for `to` (what is currently proposed in [3a197d0](https://github.com/janet-lang/janet-lang.org/pull/336/commits/3a197d0b8e58217753f68ae54e3b5e979bdbdc23)):

> Match until patt is found without including it in the match. If the
> end of the input is reached without finding patt, the entire
> pattern does not match.

Then for `thru`, rather than:

> Match until patt is included in at the end of the match. If the
> end of the input is reached without finding patt, the entire
> pattern does not match.

I think the following would be better:

> Match until patt is found and include it in the match. If the
> end of the input is reached without finding patt, the entire
> pattern does not match.

In my view, the latest proposal in [3a197d0](https://github.com/janet-lang/janet-lang.org/pull/336/commits/3a197d0b8e58217753f68ae54e3b5e979bdbdc23) for `thru` has a structure that unnecessarily deviates from that of `to`.

There is an advantage to keeping the structure similar to make it clear what the differences in behavior are.

---

However, these candidates have stripped the documentation of useful hints for future readers.  I believe that the original text had the words `to` and `through` in them was valuable in helping programmers who might have had experience with text matching to understand AND be able to recall or "recreate" the names of the PEG specials.  The names of the PEG specials are what they are, I doubt these are going to be changed.

The phrases "match up to" and "match up through" are effectively idioms among programmers -- the links above mentioned in [this comment](https://github.com/janet-lang/janet-lang.org/pull/336#issuecomment-4322258339) are evidence for this.  In my opinion it is not as important if one of them happens to not be used so much (or at all) by non-programmers.  The context here is one of programming so I think it's reasonable to suppose that what is most relevant is the language used by such folks.

If you were not familiar with such phrases, you might be better served by learning these idioms so that you can understand existing documentation and communication by programmers  (at least those with certain English experience) better.

Perhaps asking for the opinions of other Janet users on Zulip might bring additional perspectives on this matter?

sogaiu:
Possibly one could fit the phrases "up to" and "up through" in while still retaining a fair bit of the existing proposal.

So:

> Match until patt is found without including it in the match (i.e. "up to"). If the end of the input is reached without finding patt, the entire pattern does not match.

and:

> Match until patt is found and include it in the match (i.e. "up through"). If the end of the input is reached without finding patt, the entire pattern does not match.

This way folks who are familiar with certain idioms can benefit in the ways mentioned above.

amano-kenji:
I think the idioms don't make sense without code examples. I don't consider myself non-programmer.

(i.e. "up to") and (i.e. "up through") are more confusing than the original documentation. If you really want to preserve the original idioms, I'd rather write

* Match until patt is found without including it in the match. In other words, match up to patt.
* Match until patt is found, and include it in the match. In other words, match up through patt.

I personally don't like the original idioms. We can do better than our precedessors who wrote "match up to" and "match up through".

sogaiu:
I prefer having examples to complement prose, but this is not the way the page is done currently [1].  Making that kind of change could result in clearer content, but it would likely be a significant undertaking.  A simple improvement might be to point to some other content that has extensive examples though.

---

> (i.e. "up to") and (i.e. "up through") are more confusing than the original documentation. 

If those bits were the only documentation I would agree, but they are proposed in the context of the previous text (which appears immediately before):

> Match until patt is found without including it in the match

and:

> Match until patt is found and include it in the match

I think the context would likely be in the mind of the reader and for folks who are familiar with the idioms, I think they would serve to clarify the situation and also help in future recall as stated before.

However, I'm not attached to those particular ideas.

---

Regarding:

> Match until patt is found without including it in the match. In other words, match up to patt.

> Match until patt is found, and include it in the match. In other words, match up through patt.

I think it would be better to make those more similar, for example like:

> Match until patt is found without including it in the match. In other words, match up to patt.

> Match until patt is found, including it in the match. In other words, match up through patt.

---

> I personally don't like the original idioms. 

That may be so, but it may also be true that other folks actually prefer them.

> We can do better than our precedessors who wrote "match up to" and "match up through".

Since other folks might prefer the originals, "better" here may be a subjective matter.

---

[1] Revisiting the page, this assertion appears to be incorrect.  There are examples.

amano-kenji:
> Since other folks might prefer the originals, "better" here may be a subjective matter.

Do you prefer the original? A few old stackoverflow comments don't represent the majority of the programming community. The majority are young programmers who didn't grow up with specific idioms that were used in the past by a small group of people.

Over many years, I almost never encountered `match up through`. `match up through` is probably a naive attempt to convey the idea.

From my perspective, these old idioms were used by people who used code examples and didn't `understand` newcomers when they wrote documentation. I understand newcomers better than they did. I'm a better psychologist.

From my perspective, my way is superior and friendly to newcomers who don't need to understand idioms. I avoid idioms. I avoid difficult words. I just use plain english.

Which one is going to be understood `exactly` in 100 years?

* Put a bullet in his brain.
* Dispose of him
* Get rid of him

`Put a bullet in his brain` is exact and doesn't involve idioms. `Dipose of him` is a cool idiom, but it can be misinterpreted. It can mean `fire him` or `put a bullet in his brain`. The precise interpretation requires context. That's why I tend to avoid idioms. In my opinion, idioms should be deleted from english. As soon as you use idioms, you have to also share context and history with other people. I want to avoid unnecessary context. Even `up to` can be misinterpreted without context.

Technical documentation should aim for precision and accuracy.

sogaiu:
I don't really want to engage with the speculative remarks being made in your comments.  I'm not sure it's productive.  It certainly takes a fair bit of time and energy to respond and I don't know that it's so helpful in the end.

I think that you keep making assertions of certain sorts without supporting evidence is counterproductive as well, (e.g. "I'm a better psychologist." or "Technical documentation should aim for precision and accuracy.").  Please stop doing that kind of thing.  That's speculation that has been stated as factual.  It doesn't feel honest to me and frankly you are inviting hostility when we should be focused on cooperating.

If you want to state that you prefer certain things, that seems fair enough, but please do not try to pass those things off as fact.

Also, although it seems worthwhile sharing your perspective, have you considered that others might have other perspectives that might also be valid?  Just because you hold some perspective and it seems valid, I don't think that rules out the possibility that others could have perspectives that are also valid in their own way.

---

I didn't remember this, but it turns out that I am the author of the documentation in question.  The commit was part of [this PR](https://github.com/janet-lang/janet-lang.org/pull/71).  From that PR I think one can tell that:

1. There were two other people who approved the content (pepe and bakpakin),
2. pepe suggested I check the relevant commit messages for hints regarding the construction of prose (and I think I did)

The [commit for `to` and `thru`](https://github.com/janet-lang/janet/commit/6d096551f05b5a1bab300f3fb80e1c9193d7f1f6) has this content (slightly edited for spelling):

> Inpsired by the REBOL operators of the same name, these combinators match bytes up to or including a given pattern. (to patt) is (almost) equalivalent to (any (if-not patt 1)), and (thru patt) is equivalent to (* (to patt) patt). The one difference is that if the end of the input is reached and patt is not matched, the entire pattern does not match.

So although I'm the author of the text being discussed in this PR, the names `to` and `thru` were chosen by bakpakin [1].  Further, he mentions they are from REBOL [2].

Please note also that this commit message's text uses the phrase:

> match bytes up to or including a given pattern

To my mind, that contains essentially the same construct as "match up to".  I think it's likely that's part of what motivated the current text (though TBH I don't remember the details).

---

I didn't see any response to the following suggestions from [this comment](https://github.com/janet-lang/janet-lang.org/pull/336#issuecomment-4331415187):

> Match until patt is found without including it in the match. In other words, match up to patt.

> Match until patt is found, including it in the match. In other words, match up through patt.

I think these are pretty close to your latest suggestions from [this comment](https://github.com/janet-lang/janet-lang.org/pull/336#issuecomment-4326203944).

Are these not an improvement over the original for you?

Although it's been more than 5 years since the commit, I still find the current text to work for me, but I'm also ok with the proposal I put above.

One could go slightly further by using double quotes:

> Match until patt is found without including it in the match. In other words, match up "to" patt.

> Match until patt is found, including it in the match. In other words, match up "through" patt.

The connection with the names `to` and `thru` is more obvious so that might be worth doing.

---

[1] My impression having used and studied Janet, its source code, and documentation over the years is that bakpakin in some fashion acknowledges the earlier work of others by often choosing the names of things from or related to sources of inspiration.

I see this as a form of respect and it can be a hint sometimes when you want to learn more about where something came from (another example is `cmt`, which IIUC comes from Lua's `lpeg` library).

It can have the drawbacks of chosen names seeming odd, unintuitive, and/or harder to learn, but there are also advantages including:

1. relieves the burden of choosing other names (which in turn saves time and allows time and resources to be allocated elsewhere),
2. shows respect for the sources,
3. can give hints when trying to learn more (because of the specific words chosen)

I won't claim to agree with the overall philosophy, but Janet is bakpakin's project (generously shared with us all and he often accepts suggestions too).

[2] See the [Skipping Input](http://rebol.com/docs/core23/rebolcore-15.html#section-4) section of chapter 15 of Rebol/Core Users Guide.

amano-kenji:
If we strictly stick to the facts, then I remembered that `match up to patt` and `match up through patt` were the exact idioms that confused me and led us to this pull request. To beginners, these idioms won't make sense without code examples and maybe even some ASCII art.

I think we should avoid the idioms that led us here unless there is a good engineering reason to keep them. These idioms didn't paint a picture in my mind.

Here's my current suggestion.

* to: match until patt is found without including it in the match. If the end of the input is reached without finding patt, the entire pattern does not match.
* thru: match until patt is found, and include it in the match. If the end of the input is reached without finding patt, the entire pattern does not match.

If this doesn't make sense for any reason, you can make more suggestions.

sogaiu:
> If we strictly stick to the facts

I appreciate your effort on this front, thank you.

> I remembered that match up to patt and match up through patt were the exact idioms that confused me and led us to this pull request. 

Hmm, ok.  That sounds a bit different from what I remember in [this comment](https://github.com/janet-lang/janet/issues/1742#issuecomment-4318186569):

> What does "match up" mean exactly? 

That might be additional confusion though so fair enough.

For:

> To beginners, these idioms won't make sense without code examples and maybe even some ASCII art.

Perhaps you didn't mean "all beginners" here, but rather "certain beginners".  I'm going to assume that sort of interpretation.

I agree that it's likely some folks with less than a certain amount of English and/or programming experience (or those with some gaps) might find the phrases confusing [1].

Regarding:

> to: match until patt is found without including it in the match. If the end of the input is reached without finding patt, the entire pattern does not match.

> thru: match until patt is found, and include it in the match. If the end of the input is reached without finding patt, the entire pattern does not match.

I still think making the structure of both more similar is an improvement (as mentioned in [this comment](https://github.com/janet-lang/janet-lang.org/pull/336#issuecomment-4331415187)):

"to":

> Match until patt is found without including it in the match.  If the end of the input is reached without finding patt, the entire pattern does not match.

"thru":

> Match until patt is found, including it in the match.  If the end of the input is reached without finding patt, the entire pattern does not match.

How does this pair of sentences seem to you?

---

[1] I spent some time searching and investigating use of "to" and "through" yesterday and my conclusion is that often "through" seems to be more precise and "to" appears to have a wider meaning.

I tried to think of some sentences in ordinary English that mirrored the use in the documentation.

One pair that seemed close was:

* I climbed up to a hole in the ceiling.

* I climbed up through a hole in the ceiling.

I think one could argue though that the use of "up" here isn't quite the same as in the documentation.

Another pair was:

* I walked up to the doorway.

* I walked through the doorway.

Note that there is no "up" in the second sentence.  I think adding it there would result in a strange sentence unless there was some context that necessitated the use of "up" for some other reasons.

Possibly closer to the text matching might be a scene with someone describing a reading assignment, e.g.

* Please read up to chapter 8.

* Please read up through chapter 8.

This is perhaps the closest pair and might reflect something about text matching's background (possibly academic and/or research contexts).  I think without further context or clarification, the first sentence is ambiguous though.

amano-kenji:
In my mind, prices go up or down. `up` has a directional meaning. A PEG match never goes up. It only goes right.

`He is so strong that he can practically go right through the wall` means he doesn't stop at the wall but goes beyond the wall. `right` here doesn't have a directional meaning. `through` can mean `beyond`.

I'd like to avoid words like `up`, `to`, `right`, and `through` because they have a lot of connotations. The basic words have too many connotations. In english, the basic words are the most difficult words. The fundamental words are the most difficult words in human langauges because they are used so often and thus have accumulated so many connotations. This is why experts in any field use jargon instead of fundamental words to convey ideas precisely. Jargon doesn't have connotations.

Regarding

> Match until patt is found, including it in the match.

I am not fond of `, including it in the match`. I can understand this, but I never quite fully understood why this grammar exists and how it should exactly be interpreted. The interpretation of this grammar might change in the future because it doesn't seem like a precise grammar. I feel that this grammar has connotations. It is a very loose flexible grammar. That's why I replaced it with `, and ...` which feels more precise. I don't think people are going to interpret this in a different way in the future.

sogaiu:
Sorry, I edited the end of [my last comment](https://github.com/janet-lang/janet-lang.org/pull/336#issuecomment-4339974334).

Please have a look at the tail end.  It might help with the origin of "up".  I'll include the bits here anyway:

> Possibly closer to the text matching might be a scene with someone describing a reading assignment, e.g.
>
> * Please read up to chapter 8.
>
> * Please read up through chapter 8.
>
> This is perhaps the closest pair and might reflect something about text matching's background (possibly academic and/or research contexts). I think without further context or clarification, the first sentence is ambiguous though.

The short of it is, that "up" can be used in the context of reading things to mean something other than directionally upwards in some sense.  Although one might still say that "up" is upwards toward a limit.

I suppose you don't have the time for this sort of thing but I'll mention that the book "Surfaces and Essences: Analogy as the Fuel and Fire of Thinking" by Douglas Hofstadter and Emmanuel Sander was really helpful for me in thinking about this kind of situation.

If you haven't studied contemporary linguistics I also recommend exposure at some point.  It completely changed my uninformed views on language.  I realize in retrospect how difficult I had been making my own life confused by misinformed teachers earlier in my education (e.g. on matters such as grammar).

amano-kenji:
My brain crashed when it read `match up to` because I couldn't calculate the right connotations. I can interpret `up to` correctly in some word combinations, but `match up to` somehow didn't compute for me. My brain only learned `go up to` and a few other word combinations involving `up to`. It didn't learn `match up to` specifically in the context of PEG. `match up to` can have different meanings in different contexts. The interpretation of fundamental words are dependent on contexts. Communication is difficult because of context and connotations.

Context is cultural. I don't share the same culture with older programmers, younger programmers, and programmers who read diffrent books. It's impossible for one human to learn all the cultures. Programming is not one culture.

If possible, I would like to remove vernaculars and dialects and focus on not assuming context.

I can smell contexts and connotations.

sogaiu:
I'm not going to go point by point through all of what you have written.  I will say that I detect speculation stated as fact again though.

In order to keep this PR easier to focus on for us and future readers, I suggest we refrain from discussing such things here though.

---

Regarding:

> I am not fond of `, including it in the match`. I can understand this, but I never quite fully understood why this grammar exists and how it should exactly be interpreted

I don't claim to have an explanation that you will find satisfactory [1].

How about this alternative then:

"to":

> Match until patt is found (though not including it in the match). If the end of the input is reached without finding patt, the entire pattern does not match.

"thru":

> Match until patt is found (including it in the match). If the end of the input is reached without finding patt, the entire pattern does not match.

This retains a high degree of symmetry, making the key difference apparent and removing the `,
 including it in the match` construct.

---

In case it hasn't occurred to you, any changes from this PR are not guaranteed to remain in place "forever after".  Future PRs might change things.  There is more than one person who can merge PRs to this repository and there is no guarantee that the precise wording that might get worked out here will remain stable.  For example, I didn't remember that I was the author of the documentation for `to` and `thru` and there is no guarantee that any future reviewer will do historical investigation or even if they did decide to maintain things in a certain way.

Since realizing this point some years back, I decided that a better course of action for myself was to keep my own notes.  What I find clear is not the same as what others find clear and I am free to add examples and prose as I please.  I currently keep a fair number of my Janet PEG-related information in [this repository](https://github.com/sogaiu/margaret/) and I expose a fair bit of it for handy reference at the command line via [this repository](https://github.com/sogaiu/pdoc).  For PEGs, this has worked well for me.

---

[1] If you are not familiar to some degree with contemporary lingustics, I suspect you will continue to encounter these kinds of issues.  

Based on discussion with you I feel I can identify with a lot of what you have stated as I think I used to have similar difficulties though with different languages.

Now that I have some passing familiarity with some ideas from contemporary linguistics I feel I suffer much less.

It doesn't make learning language any easier for me from the perspective of having to learn a lot of special cases but I think I don't have as many psychological difficulties and confusions as before (at least not in the area of language learning).

amano-kenji:
How about this? I'm aware that anything can change over time. In 50 years, janet may not be recognizable to me.

sogaiu:
I looked at [the latest](https://github.com/janet-lang/janet-lang.org/pull/336/commits/19b00407f6223475c1bb8bea2536438e8c056430) and I see:

```
    @tr{@td{@code`(to patt)` }
        @td{ Match until patt is found(not including it in the match). If the
             end of the input is reached without finding patt, the entire
             pattern does not match. }}

    @tr{@td{@code`(thru patt)` }
        @td{ Match until patt is found(including it in the match). If the end of
             the input is reached without finding patt, the entire pattern does
             not match. }}
```

For both items, I think we'd want a space between "found" and "(".

Other than those bits, may be we're set?

amano-kenji:
Looks better now.

sogaiu:
I don't think "cancel" is used anywhere else on the [current PEG page](https://janet-lang.org/1.41.2/docs/peg.html) for similar situations.

It seems more confusing to me than the previous alternative.

amano-kenji:
How about this?

sogaiu:
I still see a form of the word "cancel" (in the past tense) being used:

```
    @tr{@td{@code`(to patt)` }
        @td{ Match until patt is found (not including it in the match). If the
             end of the input is reached without finding patt, `(to patt)` is
             cancelled. }}

    @tr{@td{@code`(thru patt)` }
        @td{ Match until patt is found (including it in the match). If the end
             of the input is reached without finding patt, `(thru patt)` is
             cancelled. }}
```

I think the earlier version from [this comment](https://github.com/janet-lang/janet-lang.org/pull/336#issuecomment-4347884655) with the spaces you added (in https://github.com/janet-lang/janet-lang.org/commit/a819a417a6293b16ded07e5b259a86253aba4ad9) was less confusing and more consistent with the other documentation on the PEG page:

```
    @tr{@td{@code`(to patt)` }
        @td{ Match until patt is found (not including it in the match). If the
             end of the input is reached without finding patt, the entire
             pattern does not match. }}

    @tr{@td{@code`(thru patt)` }
        @td{ Match until patt is found (including it in the match). If the end of
             the input is reached without finding patt, the entire pattern does
             not match. }}
```

amano-kenji:
Is it good now?

Issue #337, opened 2026-04-26
Document `til` PEG special.


amano-kenji:
`til` was added without documentation. I don't know what it does.

sogaiu:
Some `til`-related resources follow:

* [Here](https://github.com/janet-lang/janet/pull/1528) is a link to the PR that got merged for `til`.  It has background and motivation but the idea evolved during the lifetime of the PR.  [This comment](https://github.com/janet-lang/janet/pull/1528#issuecomment-2519047604) explained the final form a bit.  The explanation in the merged PR's comment refers to `sub` [1] so familiarity with that might help a bit.  Upon revisiting the PR and its associated discussion, I didn't find clear undersanding to be forthcoming, but see the next point.

* IIUC, [it turned out](https://github.com/janet-lang/janet/pull/1528#issuecomment-2521085835) that the idea of `til` had been [discussed nearly a year before the merged PR](https://github.com/janet-lang/janet/discussions/1345) in a GH discussion.

    The discussion had the following phrase which might help provide some intuition:
    > "capture like to, advance like thru"

    I found the content of the discussion much easier to understand.

* [Here](https://github.com/janet-lang/janet/blob/d9b1d711ea1fde52ac73a82088b512a3e17bad0d/test/suite-peg.janet#L722-L755) are official test cases for `til` that demonstrate some usage.

Note that [starting around here](https://github.com/janet-lang/janet/pull/1528#issuecomment-2520184574) some of us made an attempt to get the name to be something else, but this effort was not successful.

---

[1] Luckily, `sub` is documented in [this table](https://janet-lang.org/1.41.2/docs/peg.html#Combining-patterns) (currently near the end).  Here is a [PR for it](https://github.com/janet-lang/janet/pull/1344) that has an explanation.

sogaiu:
Tried to come up with a relatively simple and illustrative example of using `til`:

```janet
(peg/match ~(sequence (til ";"
                           (capture :a+))
                      (capture :d+))
           "xyz_;123")
 # =>
 @["xyz" "123"]
```

Now to try to explain the example...

For starters, there is a string "xyz_;123" which we'd like to extract the letters and numbers from but not the underscore or semicolon.

* `(til ...)` lets us express something like: for the portion of the text that precedes the semicolon, capture any leading part that is made of letters (so not the underscore), and skip the semicolon when done.  At this point we've captured "xyz".
[...]

PR #339, opened 2026-05-03
Tweak wording consistency in PEG docs

sogaiu:
This PR is an attempt to unify the use of the word "characters" over "bytes" in the documentation for each of the PEG specials.

Although "byte(s)" is used elsewhere in the prose, none of the other documentation for the other PEG specials was using the word "bytes".

Also, the way "bytes" is used appears to only be in the context of explaining the term "character":

> A character in Janet is considered a byte, so PEGs will work on any string of bytes. No special meaning is given to the 0 byte, or the string terminator as in many languages.

This inconsistency was noted in [one of #337's  comments](https://github.com/janet-lang/janet-lang.org/issues/337#issuecomment-4358643561) during a discussion.

### DIFF
--- a/content/docs/peg.mdz
+++ b/content/docs/peg.mdz
@@ -213,11 +213,11 @@ compiled to bytecode).
 
     @tr{@td{@code`(sub window-patt patt)` }
         @td{ Match @code`window-patt`, and if it succeeds match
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

Issue #341, opened 2026-05-04
Various PEG specials lack examples


sogaiu:
In recent discussions about the documentation for PEGs on the janet-lang.org website [the topic of example code came up briefly](https://github.com/janet-lang/janet-lang.org/pull/336#issuecomment-4331415187).

Looking over the [current version of the page](https://janet-lang.org/1.41.2/docs/peg.html), it appears there are some:

* the ip-address example
* the primitive patterns examples
* the grammars and recursion examples
* the string searching and other idioms examples

I think it might be worth considering some simple (illustrative) examples for:

1. `0`, `-1`, `true`, and `false` for the primitive patterns
2. at least one of each for the combining patterns
3. at least one of each for the captures
4. at least one of each for the built-in patterns

In some cases it may be difficult to come up with a simple meaningful / illustrative example (e.g. `unref`) but perhaps making an attempt will inform us better about which those might be.

I've made an attempt at 1 and am working on 2.

We might also consider linking to some other resources.  One obvious place is in [one of the files of the official tests for janet itself](https://github.com/janet-lang/janet/blob/27807d4456465bfec55aecd096822155402ecff7/test/suite-peg.janet).

---

If captures were introduced sooner in the prose I think it might make it easier to produce nicer examples, assuming we don't want examples to make use of certain constructs before they have been introduced.

FWIW, there is already a slight amount of order-inconsistency because the combining pattern section has a number of specials that refer to capturing even though that topic is not really introduced until the following section.  Specific specials for which this is the case include:

* `if`
* `if-not`
* `not`
* `backmatch`

For the first three, possibly one might view the situation as minor, but for `backmatch`, I think for meaningful example creation, it would make sense to have already been introduced to capturing (including the tagged variety).

Not sure there's an optimal arrangement but thought I'd jot down some thoughts.

sogaiu:
I'm using the following code to extract the code examples and verify correctness.

Probably want to be careful of the `peg.mdz` code example content as this checker uses `eval-string`.

```janet
# check the examples in content/docs/peg.mdz
#
# assumes:
#
# * a certain structure for the file (prose, codeblock, prose, etc.)
# * code examples are a certain way:
#   * each peg/match form on own line followed by expected value
#   * no blank lines within forms
#   * defs and defns are separated from other code by 2 newlines
#   * comment lines exist
#   * could be one big scope

(def src (slurp "content/docs/peg.mdz"))

(def codeblocks
[...]

PR #344, opened 2026-05-05
Add capture peg examples

sogaiu:
This is an attempt to partially address https://github.com/janet-lang/janet-lang.org/issues/341.

Examples for the following capture PEG specials have been added:

* `capture`, `<-`, `'`
* `backref`
* `group`
* `replace`, `/`
* `constant`
* `argument`
* `position`, `$`
* `column`
* `line`
* `accumulate`, `%`
* `cmt`
* `unref`
* `error`
* `drop`
* `only-tags`
* `nth`
* `uint`
* `uint-be`
* `int`
* `int-be`
* `number`
* `lenprefix`

**Update**: The examples should at least be "correct" as I've checked them with the script posted [here](https://github.com/janet-lang/janet-lang.org/issues/341#issuecomment-4371937255).

### DIFF
--- a/content/docs/peg.mdz
+++ b/content/docs/peg.mdz
@@ -396,6 +396,14 @@ grammars simpler.
          allowed in Janet files. To tag the capture without forcing a
          particular base, use @code`(number patt nil :tag)`. }}}
 
+@codeblock[janet]```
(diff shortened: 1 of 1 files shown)


bakpakin:
Nice, thank you! these look good.

PR #355, opened 2026-07-25
Add tuple C API docs

HParker:
Add documentation on how to work with tuples from the C API

I put it right after `arrays` in the C docs since they seem similar.

I am fairly new to Janet, so let me know if this is not the recommended way to interact with tuples. This seems to be the common pattern from inside Janet itself.

### DIFF
--- a/content/capi/buffer.mdz
+++ b/content/capi/buffer.mdz
@@ -1,6 +1,6 @@
 {:title "Buffer C API"
  :template "docpage.html"
- :order 5}
+ :order 6}
 ---
 
 ## Definition
--- a/content/capi/fiber.mdz
+++ b/content/capi/fiber.mdz
@@ -1,6 +1,6 @@
 {:title "Fiber C API"
  :template "docpage.html"
- :order 7}
+ :order 8}
 ---
 
 ## Definition
--- a/content/capi/memory-model.mdz
+++ b/content/capi/memory-model.mdz
@@ -1,7 +1,7 @@
 {:title "Janet's Memory Model"
  :nav-title "Memory Model"
  :template "docpage.html"
- :order 8}
+ :order 9}
 ---
 
 In order to write a functional and correct C extension to Janet, it
--- a/content/capi/table.mdz
+++ b/content/capi/table.mdz
@@ -1,6 +1,6 @@
 {:title "Table C API"
  :template "docpage.html"
- :order 6}
+ :order 7}
 ---
 
 ## Definition
--- a/content/capi/tuple.mdz
+++ b/content/capi/tuple.mdz
@@ -1,0 +1,98 @@
+{:title "Tuple C API"
+ :template "docpage.html"
+ :order 5}
+---
+
+The Janet tuple data structure is exposed via the @code`JanetTupleHead` struct.
+Usually, you can interact with tuples without interacting with
+@code`JanetTupleHead` directly. However, interacting with a janet tuple head can be
+helpful when embedding in non-C languages.
+Just like arrays, it is recommended to use the provided API functions.
+
+## Definition
+
+@codeblock[c]```
+/* Prefix for a tuple */
+struct JanetTupleHead {
+    JanetGCObject gc;
+    int32_t length;
+    int32_t hash;
+    int32_t sm_line;
+    int32_t sm_column;
+    const Janet data[];
+};
+typedef struct JanetTupleHead JanetTupleHead;
+```
+
+## Creating a Tuple
+
+You can create a tuple using @code`janet_tuple_begin` and @code`janet_tuple_end`:
+
+@codeblock[c]```
+Janet *dimensions = janet_tuple_begin(2);
+dimensions[0] = janet_wrap_integer(400);
+dimensions[1] = janet_wrap_integer(600);
+Janet tuple = janet_wrap_tuple(janet_tuple_end(dimensions));
+```
+
+## Accessing Tuple Values
+
+From C, you can access tuple values from a tuple passed to a C function using array indexing:
+
+@codeblock[c]```
+static Janet cfun_color_tuple(int32_t argc, Janet *argv) {
+    janet_fixarity(argc, 1);
+
+    if (janet_checktype(argv[0], JANET_TUPLE)) {
+        JanetTuple color_tup = janet_gettuple(argv, 0);
+        if (janet_tuple_length(color_tup) == 4) {
+            int r = janet_unwrap_integer(color_tup[0]);
+            int g = janet_unwrap_integer(color_tup[1]);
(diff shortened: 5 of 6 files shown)


sogaiu:
Thanks for this effort.

I don't know if there's a recommended way -- if someone else knows, I hope they'll speak up :sweat_smile: 

Below are some minor remarks:

0. I think the phrase "tuple head" might be slightly confusing so may be it's better to stick with `JanetTupleHead`.

1. Since the "Defintion" code sections for the Array, Buffer, and Table pages all maintained the comment before the first non-comment line, may be it's good to follow suit?  I think the comment is:
    ```c
    /* Prefix for a tuple */
    ```

2. May be the leading sentence could start like:

   ```
   The Janet tuple data structure is exposed via the @code`JanetTupleHead` struct.
   ```

3. May be the trailing sentence in the same paragraph could be:

   ```
   Just like arrays, it is recommended to use the provided API functions.
   ```

4. For the "Creating a Tuple" sentence, how about adding a colon at the end:

   ```
   You can create a tuple using @code`janet_tuple_begin` and @code`janet_tuple_end`:
   ```

5. Similarly for the "Accessing Tuple Values" sentence:

   ```
   From C, you can access tuple values from a tuple passed to a C function using array indexing:
   ```

6. The code in "Accessing Tuple Values" does not have a return statement so may be it's better to show some kind of elision (e.g. `...`) in an appropriate location to make it clear the code is not a full definition.

7. I found:

    > This will return memory which should be filled with your values. 

   slightly ambiguous.  How about the following instead:

    > This will return memory which should subsequently be filled with your values. 

8. May be we can use `@code` here for `n`:

   ```
   Build a tuple with @code`n` values.
   ```

9. Possibly it makes sense to mention [`janet_tuple_length`](https://github.com/janet-lang/janet/blob/c0b32d4e3798d0ceaf4131e642e59602a31ce151/src/include/janet.h#L1792) as well?

HParker:
Thanks! I think I incorporated all your improvements. I think the doc is much improved thank you.

For `7.` I took that language from the comment over `janet_tuple_begin`:

```
/* Create a new empty tuple of the given size. This will return memory
 * which should be filled with Janets. The memory will not be collected until
 * janet_tuple_end is called. */
```

I don't think we call values `Janets` in the docs, so I changed that and left out the part about collection. I also find this somewhat unclear. I can follow up with an update to the comment in Janet's code as well if you like.

sogaiu:
> I took that language from the comment over `janet_tuple_begin`:

Ah yeah, in the context of the source code, I think the comment makes more sense because one can examine the [associated source code](https://github.com/janet-lang/janet/blob/c0b32d4e3798d0ceaf4131e642e59602a31ce151/src/core/tuple.c#L31-L41) to disambiguate whereas for the page on the website there are fewer hints and possibly might have a higher probability of being in error.

Specifically for this case it seems clear looking at the C source that memory is not being filled whereas when we just have one line of C on the web page, we don't have much to go on.

It's true that the signature might be a hint, but one might also think that there could have been a mistranscription and that information had been left out accidentally (or perhaps an API changed...).  The C code in the source file is often compiled but the C on the webpage is not subject to as frequent a "verification" so might be more suspect.

What do you think?

---

> I don't think we call values Janets in the docs, so I changed that

This change seems fine to me.

> left out the part about collection.

This might be worth retaining though.

HParker:
Thanks! Added the collection part from the comment. I also added a note about `janet_tuple_end` that is another case where the code comment knows you can read the source and it might be useful to know why you should call `janet_tuple_end`.

sogaiu:
> it might be useful to know why you should call janet_tuple_end

I'm not sure whether this bit is important to include.  I think that the collection remark made sense because that's the kind of general thing that one would want to know when working in C.

Mentioning that it is important to call `janet_tuple_end` might be worth it though.

Alternatively, both things could be included, like:

> Finish building the tuple. Be sure to call this before using the tuple as it includes calculating the tuple hash used for equality checking.

HParker:
Updated to use your wording exactly. I could easily be convinced to it is too much detail though.

Let me know if I missed anything else, hopefully I got it all.

sogaiu:
I didn't notice earlier, but looking at:

````
@codeblock[c]```
#define janet_tuple_head(t) ((JanetTupleHead *)((char *)t - offsetof(JanetTupleHead, data)))
```

Return the @code`JanetTupleHead` from a @code`Janet`.
````

If I understand the prose here, it's saying `t` is of type `Janet`, but I think [`JanetTuple` is a typedef for `Janet*`](https://github.com/janet-lang/janet/blob/c0b32d4e3798d0ceaf4131e642e59602a31ce151/src/include/janet.h#L583):

```c
typedef const Janet *JanetTuple;
```

Shouldn't `t` be `Janet*` then?  [These lines](https://github.com/janet-lang/janet/blob/c0b32d4e3798d0ceaf4131e642e59602a31ce151/src/core/tuple.c#L101-L103) in the Janet source seem to support this idea.

Also, isn't the result of type `JanetTupleHead*` and not `JanetTupleHead`?

In addition, for:

````
@codeblock[c]```
#define janet_tuple_from_head(gcobject) ((JanetTuple)((char *)gcobject + offsetof(JanetTupleHead, data)))
```

Return the @code`Janet` from a @code`JanetTupleHead`.
````

I think a similar thing may apply.  The result looks like a `JanetTuple`, which is a `Janet*`, not a `Janet`.  Also, I think `gcobject` is also some pointer, probably `JanetTupleHead*` and not `JanetTupleHead`?

Perhaps the pointer-related points are implied?

**Comment by HParker** [comment]
Thanks! also great catch. I think I was a little sparse on details here because I haven't used these much. You can get by without converting back and forth manually which is really nice.

I updated the prose to also have a tiny bit more detail about what the `Janet*` and `JanetTupleHead*` convert between. Specifically the `Janet*` being the first element in the tuple. Let me know if I used the right words. I have been trying to follow the discussions in Zulip around better wording in documentation, but haven't fully internalized it yet.

**Comment by sogaiu** [comment]
I think we might want this:

```
The memory will not be collected until janet_tuple_end is called.
```

to have markup like:

```
The memory will not be collected until @code`janet_tuple_end` is called.
```

Also for:

```
Return the @code`JanetTupleHead*` from a @code`Janet*` pointing to the first @code`janet` in the tuple.
```

may be we want `Janet` capitalized, like:

```
Return the @code`JanetTupleHead*` from a @code`Janet*` pointing to the first @code`Janet` in the tuple.
```

HParker:
Updated! thank you! I appreciate you working through these with me.

pepe (approved):
LGTM now

sogaiu:
LGTM too.

Since the content is a bit outside of what I usually work with, I'd prefer to wait for @bakpakin's review / merge.

PR #360, opened 2026-07-28
Add wrapping C API section for _views

HParker:
This adds a short section to the C API wrapping section about using the *_view methods to access different Janet types that you often handle similarly in C.

I contemplated adding examples usage for each of these methods, but I worry it is too much detail and a little repetitive. I am happy to add some examples though if you like!

### DIFF
--- a/content/capi/wrapping.mdz
+++ b/content/capi/wrapping.mdz
@@ -66,6 +66,47 @@ double janet_unwrap_number(Janet x);
 int32_t janet_unwrap_integer(Janet x);
 ```
 
+### Viewing Collection Type Contents
+
+Janet provides some useful methods for accessing data from types with
+similar interfaces. Strings, symbols, keywords and buffers can all be
+treated as byte sequences, much like tuples and arrays as indexed types,
+and tables and structs as dictionaries. If you would like your program
+to accept indexed, byte or dictionary data, you can use these APIs to
+handle them smoothly.
+
+@codeblock[c]```
+int janet_indexed_view(Janet seq, const Janet **data, int32_t *len)
+```
+
+@code`janet_indexed_view` sets your provided @code`data` and
+@code`len` pointers to the contents of the array and the length of the
+array. It returns @code`1` if the view can be constructed and @code`0`
+if the type is invalid. This can construct a view for
+@code`JANET_TUPLE` and @code`JANET_ARRAY` types.
+
+@codeblock[c]```
+int janet_bytes_view(Janet str, const uint8_t **data, int32_t *len)
+```
+
+@code`janet_bytes_view` sets your provided @code`data` and @code`len`
+pointers to the contents of unsigned 8-bit array of bytes and
+@code`len` to the number of bytes. It returns @code`1` if the view can
+be constructed and @code`0` if the type is invalid. This can construct
+a view for @code`JANET_STRING`, @code`JANET_SYMBOL`,
+@code`JANET_KEYWORD`,  @code`JANET_BUFFER` types. It will also accept
+@code`JANET_ABSTRACT` if the abstract types @code`bytes` field is not
+NULL.
+
+@codeblock[c]```
+int janet_dictionary_view(Janet tab, const JanetKV **data, int32_t *len, int32_t *cap)
+```
+
+@code`janet_bytes_view` sets your provided @code`data` and @code`len`
+pointers to each key value pair in the dictionary and the number of
+key value pairs provided. It returns @code`1` if the view can be
+constructed and @code`0` if the type is invalid.
+
 ## Janet Types
 
 Before unwrapping a Janet value, one should check the value is of the


sogaiu:
FWIW, for:

> Strings, symbols, keywords and buffers can be
handled very similarly from your C program. Much like tuples and
arrays or tables and structs can be treated similarly. If you would
like your C program to be able to accept "indexable" things or
"dictionary-like" things

I think [this page](https://janet-lang.org/1.41.2/docs/data_structures/index.html) as well as [this code](https://github.com/janet-lang/janet/blob/c0b32d4e3798d0ceaf4131e642e59602a31ce151/src/include/janet.h#L607-L609) make a case that we can use the terms:

* "bytes type" / "bytes" (although I see "byteseq" on the page, I see "byte sequence" in three docstrings so I'm less sure what to use between these two if either)
* "indexed type" / "indexed" (i.e. not "indexable")
* "dictionary type" / "dictionary" (i.e. no "-like" needed)

Unlike some other terms, these seem pretty unambiguous to me, but if you've found differently, please elaborate as it's relevant to some other doc-related things I'm currently working on.

---

```
@code`jante_indexed_view`
```

seems like a typo.

---

Although I didn't mention anything for the use of "list" in [this line of #355](https://github.com/janet-lang/janet-lang.org/pull/355/changes#diff-29b756d725523a5d84149b896f7407a9a91cc30cb82a21123776f390ce1efd8aR80), I'm not so sure of using it in:

> pointers to the contents of the list and the length of the list. 

Will have a think.  One alternative is to say "indexed type", or "tuple" or "array".

---

Both instances of:

```
@code`jante_bytes_view`
```

seem like typos.  Perhaps in the second case what was intended was:

```
@code`janet_dictionary_view`
```

HParker:
Thanks! 

Embarrassing typo, sorry I missed that. should be fixed now.

I reworked the first paragraph to talk about how these types are similarly indexable, byte sequence or dictionaries instead of talking about the readers C program.

I also got rid of the word list and used the word array instead which I already use later in this same PR.

sogaiu:
Regarding these lines of the first paragraph:

> Strings, symbols, keywords and buffers can all be
treated as byte sequences. Much like tuples and arrays are indexable
or tables and structs are dictionaries.

It's perhaps a minor point, but the second sentence doesn't come across to me as a complete sentence [1].

May be we could do something like:

> Strings, symbols, keywords and buffers can all be
treated as byte sequences, tuples and arrays as indexable,
and tables and structs as dictionaries.

Not sure if that's an improvement.  May be there are better ideas.

---

[1] Not sure of the grammatical terminology here :sweat_smile:

pepe:
As I demonstrated many times in the history :-D, I am not a native speaker, yet I would present my opinion anyway for this case, I see the @HParker's wording more on point, yet I would combine it with yours as:

```
Strings, symbols, keywords and buffers can all be
treated as byte sequences, much like tuples and arrays as indexables
or tables and structs as dictionaries.
```

Yet, I do not have any stake in this, whatever gets the green, I am happy with. The good old coward's position :-).

sogaiu:
My impression is that "indexables" (noun) may not be a common term (although "indexable" (adjective) might be), so if we were to go in this direction, I think it might be an improvement to go with:

> Strings, symbols, keywords and buffers can all be
treated as byte sequences, much like tuples and arrays as indexed types,
and tables and structs as dictionaries.

since, as mentioned [elsewhere](https://github.com/janet-lang/janet-lang.org/pull/360#issuecomment-5101031509), [this Janet website docs page](https://janet-lang.org/1.41.2/docs/data_structures/index.html) has used the phrase "indexed type" for some years.

I'm trying to make an effort to keep the terminology consistent where it makes sense to do so.  For people who already have a good conceptual understanding of how Janet's internals work, this probably feels very minor and bike-sheddy, but part of the motivation for me to work on this stuff is to reduce the amount of guessing people may end up having to do when they are not familiar with the content (or have forgotten).

HParker:
Updated to your suggested phrasing. I appreciate your work to use consistent language.

sogaiu:
For reference, some of the recent discussions on Zulip [1] may be relevant to at least the current first paragraph candidate:

> Janet provides some useful methods for accessing data from types with
similar interfaces. Strings, symbols, keywords and buffers can all be
treated as byte sequences, much like tuples and arrays as indexed types,
and tables and structs as dictionaries. If you would like your program
to accept indexable or dictionary data, you can use these APIs to
handle them smoothly.

In the last sentence there is a phrase "indexable data" (implied) which one might take to include both byte sequences and indexed types.  AFAICT, we don't really seem to have an established accepted term for both.

On Zulip I mentioned that [I've found it seems to work ok to explicitly mention "bytes and indexed types"](https://janet.zulipchat.com/#narrow/channel/399615-general/topic/Janet.20Documentation.20Improvements/near/613998424), but it may be that a decision will be made to use a single term to cover both.

---

[1] See some messages starting [here](https://janet.zulipchat.com/#narrow/channel/399615-general/topic/Janet.20Documentation.20Improvements/near/613953140).

HParker:
You are right, I don't list all 3 types in that last sentence. I updated that sentence to include "byte" type as well.

sogaiu:
LGTM

pepe (approved):
LGTM

PR #363, opened 2026-08-05
Add examples for get

sogaiu:
This PR adds some examples for `get`.

On a related note, there is [a PR at the janet repository](https://github.com/janet-lang/janet/pull/1791) for updating the docstrings for some functions including `get`.  The examples in this PR are meant to be consistent with the updates in the other PR.

I'm making individual PRs for examples to make the review process smoother.  I think doing it this way may have the following benefits:

* makes the per-PR review and discussion effort more manageable which may in turn encourage participation as well as better attention per-PR
* avoids delaying the adding of examples for some functions / macros due to discussions over other functions / macros

### DIFF
--- a/examples/get.janet
+++ b/examples/get.janet
@@ -1,0 +1,28 @@
+(get "hello" 2) # -> 108
+(get @"hello" 2) # -> 108
+(get 'hello 1) # -> 101
+(get :hello 1) # -> 101
+(get :hello 8) # -> nil
+(get :hello 8 11) # -> 11
+
+(get [:bill :ted] 1) # -> :ted
+(get @[:ant :bee] 0) # -> :ant
+(get @[:ant :bee] 2) # -> nil
+(get @[:ant :bee] 2 :fox) # -> :fox
+
+(get {:length 20 :width 30} :width) # -> 30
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

examples/get.janet
```diff
@@ -0,0 +1,28 @@
+(get "hello" 2) # -> 108
+(get @"hello" 2) # -> 108
+(get 'hello 3) # -> 108
+(get :hello 3) # -> 108
```
sarna: To me, using 2 here and 3 there suggests that there should be *something* different between strings and buffers vs symbols and keywords. Having two "l"s in the string confuses it even further, as the result's the same with a different index.

sogaiu: It was supposed to be an attempt at humor (and a little nudge for people to pay closer attention), but I'm fine to choose something else.

sogaiu: I've changed the latter two.  How does it seem now?

sarna: Sorry 😅 Looks good to me now.

sogaiu: Thanks for reviewing!

PR #367, opened 2026-08-09
Modify looping page in various ways

sogaiu:
This PR contains some suggestions to the looping page based on some ideas developed while working on other website doc content as well as docstrings.

It includes (not exhaustively):

* More consistent use of the phrase "iterate over" (or similar) - there were some places where "over" was missing and those that seemed a bit awkward were modified [1]
* Substituting "bytes and indexed types" for "indexed data structure" - the latter phrase is more specific according to the data structure page, and given the content of the section, spelling things out seemed preferrable
* Use of the term "type" over "data structure" [2] in a number of locations - the data structure page does not mention things like fibers or abstract types at the moment, and the term "type" seems to be more inclusive and less confusing.
* Code sample indentation tweaks - minor horizontal whitespace adjustments were made for the sake of consistency
* Some text reflowing to aid in future side-by-side comparisons

---

[1] The phrase "iterate through" is also present on the page already and for the specific instances encountered it appeared fine.  Some searching among outside non-Janet resources suggested that this phrase is also used in some situations.

[2] Searching elsewhere (i.e. the wider internet) showed that some folks don't consider strings to be data structures while others do.  In the context of Janet's website docs, strings (as well as buffers, keywords, and symbols) are classified as "data structures".  The term "type" seems like it could be a less problematic choice.

### DIFF
--- a/content/docs/loop.mdz
+++ b/content/docs/loop.mdz
@@ -3,13 +3,14 @@
  :order 9}
 ---
 
-A very common and essential operation in all programming is looping. Most
-languages support looping of some kind, either with explicit loops or recursion.
-Janet supports both recursion and a primitive @code[while] loop. While recursion
-is useful in many cases, sometimes it is more convenient to use an explicit loop
-to iterate over a collection like an array.
+A very common and essential operation in all programming is
+looping. Most languages support looping of some kind, either with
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #371, opened 2026-08-09
docs/index: mention official archlinux package instead of aur ones

agorgl:
I pushed a little the archlinux maintainers and now janet is part of the official archlinux repos \o/


sogaiu:
Seems like a good idea given recent security-related events.

I haven't used Arch or related in some time though.  Would be nice to see some confirmation from others.

bakpakin:
Thanks, this looks good to me and the package seems done reasonably well.

PR #389, opened 2026-08-17
Add more examples for each

sogaiu:
This PR adds some more examples for `each` and removes an existing one [1].

While the existing examples demonstrate a common use case, they did not illustrate:

* use of destructuring
* use of fibers

---

[1] I removed one of the examples because it seemed mostly redundant.


pepe (approved):
LGTM

PR #390, opened 2026-08-17
Add more examples for filter

sogaiu:
This PR adds some more examples for `filter` and modifies the ordering / grouping a bit.

The additional examples add demonstration of usage with a dictionary in one case and a fiber in another.

Grouping and ordering was done to be more consistent with what's been done recently in other examples, i.e.:

* bytes type
* indexed type
* dictionary type
* fiber

It might be worth considering a convention as it can aid in more efficient scanning of examples in many cases.


pepe (approved):
LGTM

PR #398, opened 2026-08-19
Add more examples for accumulate

sogaiu:
This PR adds some more examples for `accumulate` and removes some existing ones.

The additions include uses with:

* bytes type (using `+` instead of `string`)
* dictionary
* fiber

Some existing examples were shortened.  I think it's easier to see in more environments and it takes slightly less time to comprehend without losing essential aspects of the original (except the reference to "Janet").

The original set of examples nicely showed comparison calls using `reduce`.  Due to feedback, I decided to remove the calls to `reduce`.  I think it's easy enough to look up `reduce`'s examples on the same page as well as type in relevant examples at `janet`'s repl for comparison.

### DIFF
--- a/examples/accumulate.janet
+++ b/examples/accumulate.janet
@@ -1,8 +1,11 @@
-(reduce + 0 [1 2 3 4]) # -> 10
-(accumulate + 0 [1 2 3 4]) # -> @[1 3 6 10]
+(accumulate string "" ["D" "a" "n"]) # -> @["D" "Da" "Dan"]
 
-(reduce + 0 []) # -> 0
+(accumulate + 0 "xyz") # -> @[120 241 363]
+
+(accumulate + 0 [1 2 3]) # -> @[1 3 6]
+
 (accumulate + 0 []) # -> @[]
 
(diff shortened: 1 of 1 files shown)


examples/accumulate.janet
```diff
@@ -1,8 +1,14 @@
-(reduce + 0 [1 2 3 4]) # -> 10
-(accumulate + 0 [1 2 3 4]) # -> @[1 3 6 10]
+(reduce string "" ["D" "a" "n"]) # -> "Dan"
```
pepe: I can understand, that reduce was here already, still I am not sure it is reasonable to show other function calls in the example for particular function. Just mine, indeed.

sogaiu: Thanks, I've seen your other comment on `accumulate2` and [responded there](https://github.com/janet-lang/janet-lang.org/pull/399#discussion_r3812213079).

I've also posted to Zulip to solicit additional feedback.

pepe (approved):
LGTM

pepe (approved):
LGTM

PR #399, opened 2026-08-19
Add more examples for accumulate2

sogaiu:
This PR adds some more examples for `accumulate2` and modifies some existing ones.

The additions include uses with:

* bytes type
* dictionary
* fiber

Some existing examples were shortened. I think it's easier to see in more environments.

The original set of examples nicely showed comparison calls using `reduce2`. I did similarly for some of the earlier additions but not for the later dictionary and fiber examples.

Also, it turns out I had failed to detect that the [original PR adding the initial examples](https://github.com/janet-lang/janet-lang.org/pull/261) had a file named in `accumulate2` and not `accumulate2.janet` (missing file extension), so the examples weren't showing up on the core api page.  I renamed the file in this PR, so once merged, the examples should be visible.

### DIFF
--- a/examples/accumulate2.janet
+++ b/examples/accumulate2.janet
@@ -1,0 +1,18 @@
+(reduce2 + "hi") # ->  209
+(accumulate2 + "hi") # -> @[104 209]
+
+(reduce2 - @"") # -> nil
+(accumulate2 - @"") # -> @[]
+
+(reduce2 + [1 2 3]) # -> 6
+(accumulate2 + [1 2 3]) # -> @[1 3 6]
+
+(reduce2 + []) # -> nil
+(accumulate2 + []) # -> @[]
+
+(reduce2 max [1 2 3 9 5]) # -> 9
+(accumulate2 max [1 2 3 9 5]) # -> @[1 2 3 9 9]
+
+(accumulate2 * {:a 1 :b 2 :c 4}) # -> @[2 8 8]
+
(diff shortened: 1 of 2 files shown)


examples/accumulate2.janet
pepe, line 1: Yea, same why show `reduce2`? It is confusing to me, to open a file with examples, only to read the first one for other function. I am sorry for an old man swearing at the clouds :-D.

sogaiu: Oh, I think the [original intent](https://github.com/janet-lang/janet-lang.org/pull/261) might have had something to do with the similarities between `reduce` + `accumulate` and `reduce2` + `accumulate2`.

`accumulate'`s docstring immediately mentions `reduce`:

> Similar to reduce, but accumulates intermediate values into an array.

So if you look at the docstring and then the examples, it kind of makes sense, right?

In `accumulate2`, there is no such mention of `reduce2` so it's probably less obvious why.

sogaiu: I'll ask for feedback from folks on Zulip.

Thanks for bringing this up (and all of your other reviewing!).

pepe (approved):
LGTM

sogaiu:
I'm planning to remove the calls to `reduce2`, but I'd like to do so after https://github.com/janet-lang/janet-lang.org/pull/398 is merged.

pepe (approved):
LGTM

PR #400, opened 2026-08-19
Add another example for reverse

sogaiu:
This PR adds another example for `reverse` showing that fibers can be used.

The order of the examples was changed somewhat to follow what's been done in some other examples [1].

---

[1] Roughly the order of:

* bytes type
* indexed type
* dictionary
* fiber


pepe (approved):
LGTM

PR #408, opened 2026-08-20
Add more examples for map

sogaiu:
This PR adds some more examples for `map` and modifies some existing comments.

Some of the changes are:

* demonstrating uses with bytes type, dictionary, and fiber values
* modifying some comments


pepe (approved):
LGTM

PR #410, opened 2026-08-20
Improve bindings page

veqqq:
The main purpose was to remove "immutable" and focus on `set`, but I also documented `var-` and `def-` in another commit.

### DIFF
--- a/content/docs/bindings.mdz
+++ b/content/docs/bindings.mdz
@@ -24,14 +24,16 @@ cannot be @code[set]. To use @code[set] to update a var, create it with @code[va
 (print myvar)
 ```)
 
-In the global scope, you can use the @code[:private] option on a def or var to
-prevent it from being exported to code that imports your current module. You can
-also add documentation to a function by passing a string to the @code`def` or
-@code`var` command.
+In the global scope, you can add the @code[:private] option to @code`def` or @code`var` to
+prevent the binding from being exported to code that imports your current module. 
+@code`def-` and @code`var-` implicitly use @code[:private].
+You can also add documentation to a function by passing a string to the @code`def` or
+@code`var` command:
 
 @codeblock[janet](```
 (def mydef :private "This will have private scope. My doc here." 123)
 (var myvar "docstring here" 321)
+(def- ydef "This string also has private scope.")
 ```)
 
 ### Scopes
--- a/content/docs/bindings.mdz
+++ b/content/docs/bindings.mdz
@@ -13,11 +13,8 @@ undefined symbols will raise an error.
 (def d (- c 100))
 ```)
 
-Bindings created with @code`def` have lexical scoping. Additionally, bindings
-created with @code`def` are immutable; they cannot be changed after definition.
-For mutable bindings, like variables in other programming languages, use the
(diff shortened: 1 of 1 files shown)


sogaiu:
I'm still struggling to understand the background for this PR.  I'm going to guess that it has something to do with [this message](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/table.2Fclone.20doesn.27t.20make.20a.20deep.20copy.3F/near/617786619).

Hopefully we can communicate about it before long.

veqqq:
Correct. There seem to be 3 things which can be immutable or not:
- name of a binding
- value of a binding
- target of a reference which is the value of a binding

Distinguishing them seems valuable:
- `set`able stuff (the topic of this PR)
- `put`/`push` etc.able stuff (core "immutable", I suggest we constrain the term to this)
- the result of `freeze` all the way down (Everyone says "deep copy" already! Currently the docstriings say e.g.: "(make it immutable) and do a deep copy, making child values also immutable also immutable")

There's very little documentation surface area for `var` and `def` since thye're built-ins. This single use of "immutable" seems confusing compared to describing the underlying behavior (whether or not `set` works)(illustrated by the shorter description.) `freeze` and its brethren are also a small grouping and "deep copy" serves them well. Finally, this lets us then explain equality/`sorted`'s behavior based on mutability (which the permitted keys to dictionaries already do).

N.b. I did consider using "constant" to describe `def`, but it's better to keep our vocabulary small. I also considered a phrasing with "`set`able" directly, but it sounded less natural.

sogaiu:
Thanks for elaborating.

More later, but wanted to mention first that I think the original text:

```
Additionally, bindings created with @code`def` are immutable
```

is saying that the binding itself cannot change (for the case in question).

As I understand it, a binding is an association between a name and a value.  I think the text is claiming that for `def`, the associated binding stays the same.

Please note the text from [here](https://janet-lang.org/docs/specials.html) [1]:

> This special form binds a value to a symbol. The symbol can be substituted for the value in subsequent expressions for the same result. A binding made by def is a constant and cannot be updated.

Regarding:

> I did consider using "constant" to describe def, but it's better to keep our vocabulary small.

I agree in general, but since the term "constant" is already being used elsewhere to describe the non-changing-ness of the binding, perhaps we can consider "constant" as an option.  I think we don't have to decide this immediately though.  Perhaps there is a better choice.

---

[1] The text is at the start of the section with name starting `(def name meta`.

veqqq:
Oops, I thought this was a draft. We talked about it a bit more, summarizing: As it seems `@code` can't go at the beginning of a line, I suggest prepending "both". As Sogaiu found "constant" elsewhere, I propose this as the clearest version:

> Both `def` and `var` create bindings with lexical scoping, but `def` makes constants which [set] cannot change. To use [set] to update a value, create a variable with [var]:

Do you agree @sogaiu or do you think it's fine as is (besides the formatting issue from `@code`?)

sogaiu:
No, I don't think we've really talked about this sufficiently.

You mentioned you were working on some other proposal that I've been waiting to look at.

It is still the case for me that:

> I'm still struggling to understand the background for this PR

I think it would be better to revert for now.

PR #424, opened 2026-08-25
Trim empty last lines from examples

sogaiu:
This PR trims empty last lines from Core API example files.

When an example file has an empty last line, the associated rendered example portion has an extra blank line at the end.  It doesn't look as nice [1] and it means less is visible on the screen at one time.

---

[1] Note the tops and bottoms of the example sections of the following: [1](https://janet-lang.org/1.41.2/api/index.html#%) and [2](https://janet-lang.org/1.41.2/api/index.html#string/split).


pepe (approved):
LGTM

PR #430, opened 2026-08-31
Add examples for table/clone

sogaiu:
This PR adds some examples for `table/clone`.

There are basically two collections of code:

1. The first bit tries to explicitly show some examples of what counts as updating a table.
2. The second bit tries to show an example of what one might think of as updating a table, but actually isn't.

The content was inspired by [this Zulip discussion](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/table.2Fclone.20doesn.27t.20make.20a.20deep.20copy.3F/near/617609043) and [this issue](https://github.com/janet-lang/janet/issues/576).

_Update_: Tweaked the whitespace a bit for readability.

### DIFF
--- a/examples/table_47clone.janet
+++ b/examples/table_47clone.janet
@@ -1,0 +1,47 @@
+# what counts as updating a table?
+
+(def t1 @{:a 1})
+(def t2 (table/clone t1))
+
+# 1. adding a key-value pair
+(put t1 :b 2) # -> @{:a 1 :b 2}
+
+# t2 unaffected
+t2 # -> @{:a 1}
+
+# 2. removing a key-value pair
+(put t1 :a nil) # -> @{:b 2}
+
+# t2 unaffected
+t2 # -> @{:a 1}
+
+# 3. changing a key-value pair's value
+(put t1 :a 0) # -> @{:a 0 :b 2}
+
+# t2 unaffected
+t2 # -> @{:a 1}
+
+
+# what isn't updating a table?
+
+(def arr @[:x])
+# tables are a collection of key-value pairs
+(def t1 @{:a arr})
+(def t2 (table/clone t1))
+
+# t1 and t2 have key-value pairs with values that are the same
+(= (get t1 :a) (get t2 :a)) # -> true
+
+# the key-value pairs' values are all the same array
+(= (get t1 :a) (get t2 :a) arr) # -> true
+
+# this does not count as updating t1 or t2
+(array/push arr :y) # -> arr
+
+# the value of each table's key-value pair is still arr, i.e. the
+# tables have not been updated.
+(= (get t1 :a) (get t2 :a) arr) # -> true
+
+# each table's key-value pair is arr, which now has another element
+(get t1 :a) # -> @[:x :y]
+(get t2 :a) # -> @[:x :y]


pepe (approved):
LGTM <3

pepe (approved):
Airy LGTM too

brendanhowell:
The affirmative examples in the first half are great, very clear.

But the examples in the second half might be easier to understand if we show the actual data instead of doing comparisons to prove it.  It might be simpler to just show that by changing the values in arr, the "deep" contents of t1 and t2 also get changed.

sogaiu:
Originally, I tried to show the values that would be in the array (to be more concrete -- so there would be less guessing on the reader's part), but I felt my attempt did not leave the interpretation unambiguous because just seeing values that happen to be in the array doesn't imply that a particular array is what is there.

Perhaps I can do both though.

Will make another attempt.

sogaiu:
@brendanhowell How does the latest version look?

I intentionally kept the "proof" that both tables have a key-value pair such that the value portion is `arr`.  I placed the "proof" before revealing the content of each of the key-value pairs' values.

I did this because I felt that revealing the content first could be more confusing.  That could just be me though :sweat_smile:

examples/table_47clone.janet
```diff
@@ -0,0 +1,47 @@
...
+(put t1 :a 0) # -> @{:a 0 :b 2}
+
+# t2 unaffected
+t2 # -> @{:a 1}
+
+
+# what isn't updating a table?
+
+(def arr @[:x])
+# tables are a collection of key-value pairs
```
brendanhowell, line 28: maybe rephrase this. I think for people not familiar with pointers or assembly language, this might seem confusing.  something like "with more complex data structures like tables and arrays, the reference still point to the original data."

examples/table_47clone.janet
```diff
@@ -0,0 +1,47 @@
...
+(= (get t1 :a) (get t2 :a) arr) # -> true
+
+# this does not count as updating t1 or t2
+(array/push arr :y) # -> arr
+
+# the value of each table's key-value pair is still arr, i.e. the
+# tables have not been updated.
+(= (get t1 :a) (get t2 :a) arr) # -> true
+
+# each table's key-value pair is arr, which now has another element
```
brendanhowell, line 45: I'm sorry if it sounds like hair-splitting but using the term "another" could be ambiguous or confusing when read as "an other". We need to make sure it's clear that they refer to the same data. How about "In both tables the key :a refers not to a copy but the original arr which now has an additional element."

examples/table_47clone.janet
sogaiu: >  I think for people not familiar with pointers or assembly language, this might seem confusing.

Sorry, you've lost me here.  Where do you see any reference to pointers or assembly language?

examples/table_47clone.janet
sogaiu: > I'm sorry if it sounds like hair-splitting

Not a problem.  I appreciate the feedback!

> but using the term "another" could be ambiguous or confusing when read as "an other". 

So if I understand correctly, the claim is that somehow a reader might interpret "another element" here as "a different element".  I guess it doesn't seem likely to me that someone might read it that way if they were following the code that came before, but I suppose it's possible someone might not.

I'll make a change to try to address the point.

---

Looks like things got merged, but that doesn't prevent us from making [a follow-up PR](https://github.com/janet-lang/janet-lang.org/pull/432).

examples/table_47clone.janet
sogaiu: To give some background, I very deliberately structured the second section code to focus on the fact that it's the key-value pairs that is what the table consists of [1].

I believe that not focusing on this fact can easily lead people to be confused about what counts as updating a table (cf. the [linked Zulip discussion](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/table.2Fclone.20doesn.27t.20make.20a.20deep.20copy.3F/near/617609043)).

Thus this comment:

> \# tables are a collection of key-value pairs

is a crucial observation to make so that people are nudged toward thinking about the key-value pairs themselves over what lies within / deeper a key or a value (from the key-value pair).  In the second code section here, my hope was that people would focus on `arr` itself over the elements of `arr`.

It's another reason why I backed off originally from focusing on the content of the array in the context of the tables.  Showing such content too early makes it very easy to focus on a potentially misleading part of the picture and my sense is that this is one of the things that was causing confusion when observing the Zulip discussion.

In its current form, this approach might be causing other issues though so may be those can be addressed or possibly a better approach can be arrived at.

May be saying "key-value pairs" isn't working so well...I suppose we could instead focus on the keys and values.

---

Possibly you are already familiar with the following, but FWIW (and for future readers), I should also mention that from my perspective, the examples in this PR (as well as the docstring) are meant as supplemental material to the prose of the website docs (which I am treating as background material).

We are working on possibly adding a section about "cloning" / "copying", but one of our participants is a bit lacking in time ATM, so that has paused.  For reference, one of the drafts is in [this message](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/table.2Fclone.20doesn.27t.20make.20a.20deep.20copy.3F/near/617881266).

In any case, I appreciate the discussion, so let's continue to try to improve what's here.

---

[1] Ignoring prototypes for now...also you might have seen it already, but [this](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/table.2Fclone.20doesn.27t.20make.20a.20deep.20copy.3F/near/618923135) is an earlier bit from Zulip that is along the lines of this second section of code.

sogaiu: Looks like things got merged, but that doesn't prevent us from making [a follow-up PR](https://github.com/janet-lang/janet-lang.org/pull/432).

PR #433, opened 2026-09-01
Modify first section of bindings page

sogaiu:
This PR is a follow-up to #410.

It incorporates the original's ideas [1] of:

* Avoid using the term "immutable" to describe bindings [2] and use the term "constant" instead.
* Mention `def-` and `var-`.
* Show an example for `def-`.

In addition some of the other changes made were:

* Be more explicit about what a binding is (i.e. an association - it's a "compound" entity with two pieces).
* Fill out the examples a bit more with comments and evaluation results and/or errors.
* For the term "mutable", reword to use "bindings that can be altered" [1].
* Point out that "def" and "var" are sometimes used to refer to the created bindings themselves (not necessarily the forms that create the bindings).
* Mention when one might want to use `:private` over `def-` / `var-`.
* Also show an example for `var-`.

On the whole, I tried to maintain points and ideas from the original text (possibly in changed form) rather than complete replacement.

---

[1] Thanks to @veqqq!

[2] We are thinking to try to use "mutable" and "immutable" in the context of values but not bindings.

### DIFF
--- a/content/docs/bindings.mdz
+++ b/content/docs/bindings.mdz
@@ -3,38 +3,63 @@
  :order 5}
 ---
 
-Values can be bound to symbols for later use using the keyword @code[def]. Using
-undefined symbols will raise an error.
+The @code[def] special form binds a symbol for later use.  This kind
+of association between a symbol and a value is called a binding.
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

Issue #431, opened 2026-09-01
Improve scope and declaration text


yohannd1:
As per [discussion](https://janet.zulipchat.com/#narrow/channel/399615-general/topic/Definitions.20in.20expressions/with/620225790), it would be nice to go more in depth how scope works in Janet.

1. Mention the special forms that create new scopes:

> The only special forms that introduce new binding scopes are fn, do, if, and while.

2. Talk about the scope of variable definitions in conditions of `if`/`while`:

```janet
(if (def x ...)
  (assert (truthy? x))
  (assert (falsey? x)))
```

3. Talk about the scope of nested variable definitions:

```janet
(def x (def y ...))
(assert (= x y))
```

One suggestion for a location where this an be clarified is in the [Scopes](https://janet-lang.org/1.41.2/docs/bindings.html#Scopes) section itself on the bindings page in the language documentation.

sogaiu:
I've started working on a draft and hope to have something to share soon.

sogaiu:
@yohannd1 Since #434 was merged, may be this can be closed?

yohannd1:
Sure! Will close it.

PR #434, opened 2026-09-01
Cover more material in scopes section

sogaiu:
This PR is an attempt to address #431.

I think the points mentioned in the issue were covered.  Please let me know if I missed something.

In addition, the suggested changes include:

* Surfacing of "shadowing" a bit more in prose and in examples
* Addition of some examples
* Shortening some examples (slightly compensating for the additional text that makes the overall page longer)
* Updating some comments in some examples (e.g. IIUC, some error messages have changed since the original writing)
* Some rewording in a few places

---

Some specific points regarding mutation / removal of the original text:

1. I replaced the sentence:

    > Scopes are useful for organizing your bindings and they can expand your programs.

    with:

    > Scopes are useful for organizing your bindings and can provide flexibility in expressing your programs.

    I hope that captured the spirit.  I didn't really get what "expand" meant in the original :sweat_smile: 

2. I removed a portion that recommends the use of `let` over that of `do` and `def` for at least two reasons:

    1. The existing text recommends `let` and then goes on to say that using `do` and `def` is fine, so it seemed that just not making the recommendation communicates almost the same idea...and that's shorter!
    2. In practice, I think there are some Janet-using folks [1] who actually prefer to use `def` over `let` in some situations [2].

---

[1] Disclosure: it's true that I'm one of those people...and you're going to have to work to convince me otherwise :)

[2] It might be of interest to eventually collect together more specifics on when to use one approach over the other, but I think this PR is too narrow to contain it :P

### DIFF
--- a/content/docs/bindings.mdz
+++ b/content/docs/bindings.mdz
@@ -39,15 +39,16 @@ also add documentation to a function by passing a string to the @code`def` or
 
 ### Scopes
 
-Defs and vars (collectively known as bindings) live inside what is called a
-scope. A scope is simply where the bindings are valid. If a binding is
-referenced outside of its scope, the compiler will throw an error. Scopes are
-useful for organizing your bindings and they can expand your programs.  There
-are two main ways to create a scope in Janet.
+Defs and vars (i.e. bindings) live inside what is called a scope.  A
+scope is simply where the bindings are valid.  If a binding is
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM ❤️

yohannd1:
Looks really nice as well imo! Thanks!

Issue #366, opened 2026-08-09
Improve surfacing of "surprising" aspects of comparison


sogaiu:
This is a follow-up to https://github.com/janet-lang/janet/issues/1785, where as part of investigation, the idea was floated that perhaps the documentation could be improved in some fashion.

[This comment](https://github.com/janet-lang/janet/issues/1785#issuecomment-5151025255) has a concrete suggestion, which roughly speaking, amounts to gathering together two aspects of comparing values that some might find surprising.  The aspects are more or less both already mentioned on the page, but the second one is currently in a subsequent section.

1. these operators compare values of different types in a way that the user might find surprising.

2. The primitive = operator does not compare the contents of mutable structures (arrays, tables, and buffers). Instead, it checks if the two values are the same object.

Note that the second item is incomplete in its description as the three mutable "structures" mentioned are not the only relevant ones.  See [the aforementioned comment](https://github.com/janet-lang/janet/issues/1785#issuecomment-5151025255) for further details.

---

It's possible there were other aspects that could use improvement but as I received little feedback in the form of prose, for the moment, I'll just make a note of the comment above and work toward adjusting things in a way that makes sense to me.

PR #435, opened 2026-09-02
Improve comparisons page

sogaiu:
This PR is an attempt to address #366.

The main suggested change is to gather together the two "surprising" characteristics of how `=` compares types.

Some additional changes include:

* Spell out a bit more about how `=` behaves.  Specifically, add mention of fibers and some abstract types and fill out corresponding examples a bit.
* Replace uses of `assert` with the `# -> <return value>` style of annotating evaluation results to align with most of the rest of the website prose.
* Modify some examples to not use `not=` and `deep-not=` [1] and use instead `=` and `deep=`.

---

[1] Although all instances of `not=` and `deep-not=` were removed from examples, these were not described in the prose originally (nor in the code comments), so it seemed acceptable to do.

### DIFF
--- a/content/docs/comparison.mdz
+++ b/content/docs/comparison.mdz
@@ -16,12 +16,18 @@ as follows: @code[(< a b)] will return true if a < b, and false otherwise.
 Similarly @code[(= a b)] will be true only if a == b.
 
 @codeblock[janet](```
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #440, opened 2026-09-06
Add example for thaw

sogaiu:
This PR adds an example for `thaw` that demonstrates a resulting key being mutable.

It's intended to complement an example for `thaw-keep-keys` (in #441) that demonstrates a key being left alone in the return value.


pepe (approved):
LGTM

PR #445, opened 2026-09-15
fix Makefile

whamtet:
It looks like you don't run

    make wasm

every day.  One of the command flags needs to be updated.

Great work on Janet!


sogaiu:
Possibly relevant: https://github.com/emscripten-core/emscripten/issues/13812

